PR #3, opened 2017-05-30
Add database backend and mutation to GraphQL API.

samiwel:
## Description
This PR extends the Hello World GraphQL prototype by adding a persistence solution.
Added sequelize ORM to the project, which at the moment persists values to a Postgres DB (but also supports MySQL, SQLite etc.)
Defined a data model which Sequelize understands and seed the database with some default values.
Extended the GraphQL api to wrap calls to sequelize.
Added root query and mutation objects to the GraphQL API.

## ToDo
Work still outstanding:
- ~Add docker configuration for the postgres database.~
- ~Link postgres container and nodejs container via docker-compose configuration and use environment variables for DB config.~
- ~Update the README with instructions on how to issue queries and call mutations.~


samiwel:
[Author API Backend (5)](https://trello.com/c/IlajmbOL/42-author-api-backend-5)

src/schema.js
```diff
@@ -1,13 +1,90 @@
...
+  GraphQLList,
+  GraphQLNonNull
+} = require('graphql');
+
+// Import the Database connection.
+var Db = require('./db')
+
+// Define the GraphQL types for the schema.
+// Eventually these might include things like Pages, Questions, Answers etc.
+var Message = new GraphQLObjectType({
```
hamishtaplin: The convention in JS is to use uppercase for class definitions and lowercase for instances, eg:

```js
var message = new GraphQLObjectType
```

src/schema.js
```diff
@@ -1,13 +1,90 @@
...
 * This is where all the GraphQL schema types will be defined.
 * The finished schema is likely to include things like Pages, Questions, Answers etc.
 */
-var schema = buildSchema(`
-  type Query {
-    hello: String
-  }
-`);
+var schema = new GraphQLSchema({
+  query: Query,
```
hamishtaplin: Following aforemention convention would allow you to use object shorthand notation here:

```js
var schema = new GraphQLSchema({
  query,
  mutation
})
```

:)

src/settings.js
```diff
@@ -1,15 +1,26 @@
...
-var port = process.env.PORT || 4000;
-var graphiql_enabled = process.env.GRAPHIQL_ENABLED || true;
-var graphiql_endpoint = process.env.GRAPHIQL_ENDPOINT || '/graphiql';
+var express = {
+  port: process.env.PORT || 4000,
+  url: (process.env.SCHEME || 'http://') + (process.env.HOST || 'localhost') + ':' + (process.env.PORT || 4000)
+};
+
+var graphiql = {
+  enabled: process.env.GRAPHIQL_ENABLED || true,
```
WickyNilliams: You can never turn this off because `false || true` will always evaluate to `true`

src/settings.js
```diff
@@ -1,15 +1,26 @@
...
+  endpoint: process.env.GRAPHIQL_ENDPOINT || '/graphiql',
+  pretty: process.env.PRETTY_PRINT_GRAPHQL || true
```
WickyNilliams: Same here

docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+      - DATABASE_USERNAME=postgres
+      - DATABASE_PASSWORD=mysecretpassword
+      - DB_HOST=db
+      - DB_DIALECT=postgres
+  db:
+    image: postgres
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=mysecretpassword
+      - POSTGRES_DB=postgres
```
WickyNilliams: These env vars do not seem to be referenced anywhere, are they automatically picked up by postgres?

samiwel: Yes, that's correct, they are used by the `postgres` official docker image. See **Environment Variables** section of https://hub.docker.com/_/postgres/. Over time, the values will be removed from the `docker-compose.yml` file, particularly if we deploy somewhere, they'll likely need to be injected from somewhere, but for running locally this works fine for now.

src/settings.js
samiwel: Good spot. I'll remove the defaults for Boolean environment values.

WickyNilliams:
I tested the first-time setup, and there are no longer issues with missing node dependencies 👍 

I notice that every restart of the app adds another value to the DB. Also, the current `addMessage` mutation always inserts and never updates messages. Is it worth adding some mechanism for updates as a proof of concept?

samiwel:
@WickyNilliams Added mutations for `updateMessage` and `deleteMessage`. Please would you mind checking over the delete just to make sure I'm not doing anything dumb with the Promise?

PR #5, opened 2017-06-08
45/data model

WickyNilliams:
This PR adds:

* DB tables
* GraphQL schema
* Tests

For

* Questionnaire
* Group (i.e. Section)
* Page (super-type)
* Question (sub-type of Page)
* Answer

Matching this ER diagram

author-api-er

## Todo

- [x] deletions
- [x] updates
- [x] improve file structure / modularise graphql schema
- [x] tests
- [x] add `surveyId` to `Questionnaire` model/query
- [x] update data model so that Question is a type of Page
- [x] add groups as child of Questionnaire and parent of Page
- [x] Use GraphQL enums where appropriate
- [x] Add any useful info to README
- [x] I think we should rename `Question` to `QuestionPage` throughout to make it clear that it's a type of page. This pattern should then be followed for `InterstitialPage` etc
- [x] It looks like some of the PageRepository methods won't work e.g. the `insert` method calls `create` on a `QuestionRepository` but it's the `Question` model that has a `create` method.
- [x] The models appear to be database access so should probably be moved to the db folder.
- [x] There is some inconsistent naming in the GraphQL API e.g. `GroupId` should be `groupId` to be consistent with the other properties.
- [x] We don't really have a limit on length for labels and titles so they should probably be defined as text rather than string.
- [x] The knex config is defined away from the other config - can this be rationalised so that we keep the config in the same place?


WickyNilliams:
[Create Author Questionnaire (5)](https://trello.com/c/W5HiTx8W/45-create-author-questionnaire-5)

docker-compose.yml
```diff
@@ -13,15 +13,11 @@ services:
       - 4000:4000
       - 5858:5858 # open port for debugging
     environment:
+      - DB_CONNECTION_URI=postgres://postgres:mysecretpassword@db:5432/postgres
```
samiwel: Nice tidy up 👍

graphql/schema.js
```diff
@@ -0,0 +1,270 @@
...
+          type : GraphQLBoolean
+        },
+
+        questionId: {
+          type : GraphQLID
+        }
+      },
+
+      resolve(source, { title, description, guidance, label, qCode, type, mandatory, questionId }) {
+        return models.Answer.create({
```
samiwel: Just occurred to me that maybe we should introduce a "Service" layer in between the GraphQL API and Sequelize. It may be a bit overkill for now, but it would get rid of the direct dependency on Sequalize from the GraphQL api. Another benefit of doing this would be that additional Business logic e.g authentication could be done in this service layer. We don't need it yet for this PR though, just something to consider when we need it I suppose. 👍

WickyNilliams: For isolating ourselves from sequelize, we're using the graphql-sequelize lib here to generate the schema, so we're always going to bound to it somewhat. We pay for DRY with some coupling.

As for auth, that could/would be handled by an earlier express middleware, so graphql can be completely ignorant of that concern. Resolvers also receive the request object (which would be decorated with user info by that point), so you can then select by `req.user.id` or something like that

WickyNilliams:
@hamishtaplin i think we should do that as a separate task. I'm conscious that this task has already ballooned, and I don't want it to drag on

samiwel:
@WickyNilliams Is this ready to be merged? There was some talk of changing the schema yesterday to introduce groups/question-pages, do you think it would be better to make those changes under another PR?

WickyNilliams:
@samiwel i guess that question was answered in stand up. Perhaps it can still be reviewed without worrying too much about the data model side of things? 

I'm also starting to wonder if sequelize is going to become restrictive/limiting, if our data model gets more complex. I don't think it supports "is-a" taxonomies, only "has-a".  Thankfully graphql supports  "is-a" through [unions](https://medium.com/the-graphqlhub/graphql-tour-interfaces-and-unions-7dd5be35de0d).

migrations/20170620163533_create_questionnaires_table.js
```diff
@@ -0,0 +1,22 @@
+
+exports.up = function(knex) {
```
samiwel: Are these migrations generated or did you have to hand-craft them?

schema/types/Answer.js
```diff
@@ -0,0 +1,40 @@
+const { GraphQLObjectType, GraphQLString, GraphQLBoolean, GraphQLInt } = require("graphql");
+
+module.exports = new GraphQLObjectType({
+  name : "Answer",
+  description : "An answer",
+
+  fields : () => ({
+    id : {
+      type: GraphQLInt
```
samiwel: Just a small point but I wonder if the `Id` fields should be of type `GraphQLID` for better semantics?

samiwel:
@WickyNilliams Lots going on here, but it generally looks good to me. I think the approach taken is good and the repository pattern helps with separation of concerns. Nice that you can also set the repositories on the context. Didn't realise you could do that. I'd get someone else to review in case I missed something, but otherwise I think this is good to squash and merge.

Wonder if it's worth adding a bit of documentation, either in the README or in confluence about how to query this model now that Question IS A Page?

schema/types/Answer.js
WickyNilliams: I had that originally, but removed it because graphql implicitly casts IDs to String, and it was breaking my tests. I guess now that it's (essentially) code complete I could go back and do that.

knexfile.js
```diff
@@ -0,0 +1,41 @@
+// Update with your config settings.
+
+module.exports = {
+
+  development: {
+    client: 'postgresql',
```
samiwel: As a longer-term tech debt thing, I wonder if we should re-visit this. I think we should be relying on passing these things in as environment variables rather than dictating what client we're using in a config file. i.e. What if I want to use sqlite or mysql during development?

schema/types/Answer.js
samiwel: If it causes problems with implicit casting then don't worry about it. It was only a small point and int makes sense given the underlying types of the db columns.

migrations/20170620163533_create_questionnaires_table.js
WickyNilliams: The file and skeleton of migration is generated with `yarn knex -- migrate:make name_of_migration`. The rest is handwritten.

There are [tools](https://github.com/jwilm/knex-migrate) to generate more of the migration based on naming conventions (a la Rails), but I didn't want to throw that in the mix right now.

I'll update README to include details about creating migrations

knexfile.js
WickyNilliams: I think we'll eventually hit a point where we have to use DB-specific features. I'm already using Postgres' `RETURNING` feature to avoid a second query when creating/updating/deleting (to fetch the newly added/updated/deleted record).

Another example (which I haven't done yet), is a `TRIGGER` for auto-updating the `updated_at` columns. MySQL and Postgres have different approaches to this, so the code wouldn't be portable. 

It's worth mentioning that `updated_at` stuff can be managed in code, but it's often safer/easier to do with a TRIGGER

WickyNilliams:
@samiwel 

> Lots going on here, but it generally looks good to me. I think the approach taken is good and the repository pattern helps with separation of concerns. 

Yeah it definitely helps. Once I introduced them I could migrate the data access layer to Knex without changing any consumers of repositories. Not sure about division of responsibilities still. e.g. I don't think the models should know anything about the DB, that should all be handled in a Repository. I left it like this as I wanted to mimic the Sequelize setup for a clean migration to Knex. So I've left it as-is for now (it'll be an easy refactor later)

> Nice that you can also set the repositories on the context. Didn't realise you could do that. 

Took me a while to realise too. Super handy for making things testable!

> I'd get someone else to review in case I missed something, but otherwise I think this is good to squash and merge.

Sure. The more eyes, the better. 

> Wonder if it's worth adding a bit of documentation, either in the README or in confluence about how to query this model now that Question IS A Page?

Good idea, I'll see what I can add

package.json
```diff
@@ -1,25 +1,46 @@
...
   "dependencies": {
     "colors": "^1.1.2",
     "dotenv": "^4.0.0",
+    "eslint": "^4.0.0",
+    "eslint-plugin-jest": "^20.0.3",
+    "jest": "^20.0.4",
+    "jest-environment-node-debug": "^2.0.0",
+    "nodemon": "^1.11.0",
+    "sequelize-cli": "^2.7.0",
+    "sequelize-fixtures": "^0.5.6",
```
WickyNilliams: Good catch! Removed

PR #6, opened 2017-06-22
Make changes to package.json for CloudFoundry.

samiwel:
## Description
CloudFoundry builds and runs the app, but it does not keep the dev dependencies around, so the package.json was updated so that dependencies are under the main runtime dependencies.

Also CloudFoundry needed the engines property as a hint to select the correct version of nodejs_buildpack.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,18 +6,20 @@
   "scripts": {
     "start": "nodemon --debug"
   },
-  "devDependencies": {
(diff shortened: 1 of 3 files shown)


samiwel:
[Integrate deployed CF environments (5)](https://trello.com/c/wvJErhS1/94-integrate-deployed-cf-environments-5)

PR #10, opened 2017-07-03
Added migrations to docker build

emilioward:
This PR adds the migration command to the docker file inside the scripts/run_app.sh.

To test this stop and delete the docker image and run it again to run the migration and check to see if the tables are created inside postgres.

### DIFF
--- a/scripts/run_app.sh
+++ b/scripts/run_app.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
 
 yarn install
+DB_CONNECTION_URI=postgres://postgres:mysecretpassword@db:5432/postgres yarn knex -- migrate:latest
 yarn start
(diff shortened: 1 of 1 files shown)


scripts/run_app.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 
 yarn install
+DB_CONNECTION_URI=postgres://postgres:mysecretpassword@db:5432/postgres yarn knex -- migrate:latest
```
samiwel: Just spotted that we're actually setting the environment variable in the `docker-compose.yml` file [see here](https://github.com/ONSdigital/eq-author-api/blob/48e247e8cfc31bf1cb116fc2a6ed7c722c4f523f/docker-compose.yml#L16) so we can probably just change this line to `yarn knex -- migrate:latest`. Might want to give it a test though to be sure that works.

emilioward: Ive added that change in and it works from testing, i'll push the change up

PR #15, opened 2017-07-06
Split out the GraphQL schema definition from the API

samiwel:
## Description
This PR implements the recommend best practice [as described on Apollo's website](http://dev.apollodata.com/tools/graphql-tools/generate-schema.html) by using the `graphql-tools` module to separate the GraphQL schema type definitions from the API.
This work is intended to make it easier. and pave the way for [allowing the API to be mocked](http://dev.apollodata.com/tools/graphql-tools/mocking.html) in both the Author and Publisher applications.
The change also makes a number of files redundant so they have been removed from the project to keep the codebase smaller and hopefully more maintainable going forward.

There is a further bit of work required to actually move the Schema into its own repository, so that it can be shared between the projects.
This should be done under another task.

## How to test
Everything should still work as normal.
All tests and build should pass.
Developers should be able to use the API as they did previously.


samiwel:
[Split Schema from API (3)](https://trello.com/c/f8S4LVxI/118-split-schema-from-api-3)

samiwel:
@WickyNilliams Be good just to confirm all still working between the Author -> API and Publisher->API before merging - I'm testing publisher now 👍

WickyNilliams:
@samiwel I'll test against my latest author branch now

samiwel:
Update: Seems to work okay with Publisher. Was able to use mutations using GraphiQL and query from Publisher with no problems.

WickyNilliams:
Looks good from this side

PR #18, opened 2017-07-13
make backwards compatible

WickyNilliams:
* adds `groups` back for backwards compatibility
* uses `@deprecated` directive to notify clients of deprecation (e.g. graphiql shows it is deprecated, and gives reason for deprecation)


WickyNilliams:
[Author API - Rename Groups -> Sections (2)](https://trello.com/c/xsJpEiaf/96-author-api-rename-groups-sections-2)

PR #19, opened 2017-07-14
Create initial section & page

WickyNilliams:
Creates a new section and page when a questionnaire is created. 

We realised while building the eq-author front-end that this is an implicit requirement when creating a questionnaire, and so the API should assume the responsibility rather than have the front-end make multiple sequential requests.

### DIFF
--- a/schema/resolvers.js
+++ b/schema/resolvers.js
@@ -12,8 +12,25 @@ const Resolvers = {
   },
 
   Mutation: {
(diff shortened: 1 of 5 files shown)


WickyNilliams:
[Modify API to create new section when creating new Questionnaire](https://trello.com/c/aJ69DvDJ/130-modify-api-to-create-new-section-when-creating-new-questionnaire)

PR #21, opened 2017-07-17
Prettier

bennpearson:
Added prettier. Check it runs locally and on travis


bennpearson:
[Prettier for API (0.5)](https://trello.com/c/QbgbYjFx/86-prettier-for-api-05)

PR #24, opened 2017-08-09
make deletePage work

WickyNilliams:
Allows deletion of a `Page` without having to know `pageType` (e.g. so no need to call `deleteQuestionPage`)

### How to review 
test with accompanying PR in author


WickyNilliams:
[Delete Page (5)](https://trello.com/c/h4lkTX9Y/156-delete-page-5)

PR #29, opened 2017-08-24
Change api to return objects

samiwel:
This PR removes foreign key details from leaking into the GraphQL API, by updating the resolvers to return references to the actual entities.

In practice what this allows you to do is browse the API as a graph of objects, so you can go in both directions, up and down the tree.

So for example, you can now do things like:

```
{
  questionnaire(id: 1) {
     title
     sections {
       id
       questionnaire {
          title
       }
     }
  }
}

```

in the above example, both `title` attributes should be equal.

### How to review 
This PR is not ready for merge until the schema and the front-end changes have been completed.
They will be worked on in separate PRs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "colors": "^1.1.2",
     "cors": "^2.8.3",
     "dotenv": "^4.0.0",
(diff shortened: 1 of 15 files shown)


samiwel:
[Refactor data management (8)](https://trello.com/c/AeSZn30W/191-refactor-data-management)

PR #31, opened 2017-09-11
Remove type from question

samiwel:
This PR has all of the API changes necessary to remove the type field from the QuestionPage GraphQL Schema type.
This change removes all references to `type` from queries and resolvers present in the Author API.
It also introduces a database migration script that can be used to drop the appropriate column on the Pages table.

### How to review 
All tests should pass.
There should be no references to `type` on the QuestionPage related objects.
The database migration script should behave as expected.


samiwel:
[Author API - Remove QuesitonPage.type (2)](https://trello.com/c/4QIMS7P0/97-author-api-remove-quesitonpagetype-2)

PR #38, opened 2017-09-25
Improve graphql usage

WickyNilliams:
**DO NOT MERGE UNTIL SCHEMA MERGED**

Backwards-compatible changes to use `ID` type for entity id's, and input types for all mutations

### How to review 
run tests, examine changes, test against master of `eq-author` and [associated PR](https://github.com/ONSdigital/eq-author/pull/211) (to ensure it is compatible with both)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       - db
     volumes:
       - .:/app
-#      - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
+      # - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
     ports:
       - 4000:4000
       - 5858:5858 # open port for debugging
--- a/schema/resolvers.js
+++ b/schema/resolvers.js
@@ -1,21 +1,29 @@
-const { merge, includes } = require("lodash");
-const GraphQLDate = require("graphql-iso-date").GraphQLDate;
+const { GraphQLDate } = require("graphql-iso-date");
+const { includes } = require("lodash");
+
+const getId = args => args.newId || args.id;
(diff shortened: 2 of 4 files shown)


WickyNilliams:
[Improve GraphQL usage (5)](https://trello.com/c/8bkFSzmj/187-improve-graphql-usage-5)

docker-compose.yml
```diff
@@ -9,7 +9,7 @@ services:
       - db
     volumes:
       - .:/app
-#      - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
+      - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
```
samiwel: Is it right that we want to uncomment this?

schema/resolvers.js
```diff
@@ -1,21 +1,31 @@
-const { merge, includes } = require("lodash");
-const GraphQLDate = require("graphql-iso-date").GraphQLDate;
+const { GraphQLDate } = require("graphql-iso-date");
+const { includes } = require("lodash");
+
+const getId = args => args.newId || args.id;
+const getNewId = entity => entity.id.toString(10);
+const getInputArgs = args => args.input || args;
+const getIdFromArgs = args => getInputArgs(args).id;
```
samiwel: Would there ever be a need to check for `newId` here? e.g. `const getIdFromArgs = args => getId(getInputArgs(args));`

WickyNilliams: No, should be fine. This is only used for mutations, there the named arg is called `id` and the field on the input type is also `id`

docker-compose.yml
WickyNilliams: Good catch

PR #41, opened 2017-10-20
Add structured logging of req/res and logging of GraphQL errors.

samiwel:
This PR introduces some structured machine-readable (JSON) logging to the Author API.
The `express-pino-logger` middleware function is used to log request and responses.
This PR also uses `graphql-tools` and `formatError` to ensure that any errors in the GraphQL query responses are captured.

### How to review 
All tests should pass.
Observe the JSON log format being output to the console.
Generate an error in the GraphQL query and observe the error being logged in the console (as well as in the browser).


samiwel:
[Apply logging to API (3)](https://trello.com/c/vzrzAbLs/278-apply-logging-to-api-3)

app.js
```diff
@@ -1,28 +1,44 @@
...
-
+const pino = require("express-pino-logger")();
 const { PORT } = require("./config/settings");
 
 const app = express();
-app.use(morgan("tiny"));
+
+app.use(pino);
+
+const formatError = e => {
```
WickyNilliams: This is badly named. Seems better to extract the `logger` object below, and add a `log` method to it?

app.js
```diff
@@ -1,28 +1,44 @@
...
+};
 
 app.use(
   "/graphql",
   cors(),
   bodyParser.json(),
-  graphqlExpress({ schema: schema, context: { repositories: repositories } })
+  graphqlExpress({
+    schema: withErrorLogging(schema),
+    context: { repositories: repositories },
```
WickyNilliams: `context: { repositories }`

app.js
samiwel: Only reason for naming it `formatError` was because it's passed into `graphqlExpress` options. Would something like this be better?

```
const log = error =>{
     console.error(error);
     return e;
} 
const logger = {
 log
}
```

then

`..., formatError: logger}`

PR #44, opened 2017-10-25
205 undo deletes

samiwel:
This PR makes changes to the API, namely in the repositories and DB access layer to support soft deleting of Questionnaires, Sections, Pages, Answers and Options.
A new DB migration script has been included which adds a new column `state`. For now possible values for `state` are `Active` (the default value), and `Deleted`.
In future, there may be additional states such as 'Archived' or 'Locked' but that is not in scope for now.
A new integration test has been introduced to test soft deletion at the Questionnaire level.

### How to review 
When running the author against this branch, deleting a Questionnaire, Section, Question, Answer or Option should mark the correct record as being "Deleted".
The deleted element should no longer appear when using the application. The deleted element can be restored by setting the state value back to `Active`.

### DIFF
--- a/db/Answer.js
+++ b/db/Answer.js
@@ -4,12 +4,9 @@ function Answer() {
   return db("Answers");
 }
 
-module.exports.findAll = function findAll(where = {}) {
-  return Answer()
-    .where(where)
-    .select();
+module.exports.findAll = function findAll() {
+  return Answer().select();
 };
-
 
 module.exports.findById = function findById(id) {
   return Answer()
@@ -35,4 +32,4 @@ module.exports.destroy = function destroy(id) {
     .where("id", parseInt(id, 10))
     .delete()
     .returning("*");
-}
+};
--- a/db/Questionnaire.js
+++ b/db/Questionnaire.js
@@ -4,12 +4,9 @@ function Questionnaire() {
   return db("Questionnaires");
 }
 
-module.exports.findAll = function findAll(where = {}) {
-  return Questionnaire()
-    .where(where)
-    .select();
+module.exports.findAll = function findAll() {
+  return Questionnaire().select();
 };
-
 
 module.exports.findById = function findById(id) {
   return Questionnaire()
@@ -35,4 +32,4 @@ module.exports.destroy = function destroy(id) {
     .where("id", parseInt(id, 10))
     .delete()
     .returning("*");
-}
+};
--- a/repositories/AnswerRepository.js
+++ b/repositories/AnswerRepository.js
@@ -1,21 +1,26 @@
 const { head, invert, map } = require("lodash/fp");
 const Answer = require("../db/Answer");
 const mapFields = require("../utils/mapFields");
-
 const mapping = { QuestionPageId: "questionPageId" };
 const fromDb = mapFields(mapping);
 const toDb = mapFields(invert(mapping));
 
 module.exports.findAll = function findAll(
-  where,
+  where = {},
   orderBy = "created_at",
   direction = "asc"
 ) {
-  return Answer.findAll(where).orderBy(orderBy, direction).then(map(fromDb));
+  return Answer.findAll()
+    .where({ isDeleted: false })
+    .where(where)
+    .orderBy(orderBy, direction)
+    .then(map(fromDb));
 };
 
 module.exports.get = function get(id) {
-  return Answer.findById(id).then(fromDb);
+  return Answer.findById(id)
+    .where({ isDeleted: false })
+    .then(fromDb);
 };
 
 module.exports.insert = function insert({
@@ -49,7 +54,8 @@ module.exports.update = function update({
   label,
   qCode,
   type,
-  mandatory
+  mandatory,
+  isDeleted
 }) {
   return Answer.update(id, {
     description,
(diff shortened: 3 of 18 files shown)


samiwel:
[Undo Deletes (8)](https://trello.com/c/2ch8mhgt/205-undo-deletes-8)

db/Questionnaire.js
```diff
@@ -33,6 +35,6 @@ module.exports.create = function create(obj) {
 module.exports.destroy = function destroy(id) {
   return Questionnaire()
     .where("id", parseInt(id, 10))
-    .delete()
+    .update("state", "Deleted")
```
WickyNilliams: Worth creating some constants for these?

repositories/AnswerRepository.js
```diff
@@ -8,14 +8,17 @@ const toDb = mapFields(invert(mapping));
...
 module.exports.findAll = function findAll(
   where,
+  whereNot = { state: "Deleted" },
   orderBy = "created_at",
   direction = "asc"
 ) {
-  return Answer.findAll(where).orderBy(orderBy, direction).then(map(fromDb));
+  return Answer.findAll(where, whereNot)
+    .orderBy(orderBy, direction)
+    .then(map(fromDb));
```
WickyNilliams: I've just realised, since the DB layer returns the knex object, we could actually add this functionality without touching that layer at all. This should be identical functionally:

```js
return Answer.findAll(where)
  .whereNot(whereNot)
  .orderBy(orderBy, direction)
  .then(map(fromDb));
```

I think most (all?) of your changes to the access layer can be reverted if we follow this pattern

schema/resolvers.js
```diff
@@ -3,6 +3,7 @@ const { includes } = require("lodash");
 
 const getId = args => args.newId || args.id;
 const getNewId = entity => entity.id.toString(10);
+const undelete = obj => Object.assign({}, obj, { state: "Active" });
```
WickyNilliams: I wonder whether we should just go ahead and add an undelete method to Repository objects? Might be a bad idea, thinking aloud

tests/integration/db/Questionnaire.test.js
```diff
@@ -0,0 +1,117 @@
+process.env.NODE_ENV = "test";
```
WickyNilliams: This should instead be set for the test run in package.json. e.g. 

```
"test": "NODE_ENV=test jest .",
```

tests/integration/db/Questionnaire.test.js
```diff
@@ -0,0 +1,117 @@
...
+
+const knex = require("../../../db");
+
+const migration = require("../../../migrations/20170620163533_create_questionnaires_table");
```
WickyNilliams: Why do you need to manually run the migration?

schema/resolvers.js
samiwel: Personally, I prefer to keep repositories as CRUD as possible (with exception of Find and FindAll - this is similar to how I've seen them in other frameworks. It feels like undeleting belongs at a "Service" level above the repository.

tests/integration/db/Questionnaire.test.js
samiwel: For this integration test I needed the questionnaire table in place. I originally ran all migrations on a beforeAll step (against a SQLite database). Unfortunately, some of the migrations use features available in postgres, but not in SQLite. So running all migrations is not possible. For the purposes of the test. All I care about is the existence of the table. So I'm able to run that migration in isolation.

db/Answer.js
```diff
@@ -10,16 +10,17 @@ module.exports.findAll = function findAll(where = {}) {
     .select();
 };
 
-
-module.exports.findById = function findById(id) {
+module.exports.findById = function findById(id, where = {}) {
```
WickyNilliams: remove the `where` arg (here and elsewhere) as discussed yday - any further filtering can be added at repository layer

repositories/AnswerRepository.js
```diff
@@ -1,21 +1,25 @@
...
 const mapFields = require("../utils/mapFields");
-
 const mapping = { QuestionPageId: "questionPageId" };
 const fromDb = mapFields(mapping);
 const toDb = mapFields(invert(mapping));
+const { STATE_ACTIVE, STATE_DELETED } = require("../constants/recordStates");
 
 module.exports.findAll = function findAll(
   where,
+  whereNot = { state: STATE_DELETED },
```
WickyNilliams: This should be express in the positive (`where = {state : STATE_ACTIVE }`, or `where = { isDeleted : false }`)

repositories/AnswerRepository.js
```diff
@@ -1,21 +1,24 @@
...
-  return Answer.findAll(where).orderBy(orderBy, direction).then(map(fromDb));
+  return Answer.findAll(where)
+    .whereNot(whereNot)
+    .orderBy(orderBy, direction)
+    .then(map(fromDb));
 };
 
-module.exports.get = function get(id) {
-  return Answer.findById(id).then(fromDb);
+module.exports.get = function get(id, where = { isDeleted: false }) {
```
WickyNilliams: I wonder whether it's even worth having this `where` param for `get`? You're trying to select a single item, so it doesn't make sense to select on anything other than `id` (at least from the perspective of the calling code)

PR #47, opened 2017-11-23
Add status page to the author api.

samiwel:
Adds a `/status` endpoint to the author API. The git commit hash will be displayed in the status output when built into the docker container.

### How to review 
To test this you should be able to hit the `/status` endpoint and observe that you get an OK result when the API is running.
To test displaying the GIT SHA-1 hash, you need to build the docker image passing in the arg (normally would be performed by travis).
Ensure API docker container is stopped. Build the image with the following command:
```docker build -t eqauthorapi_web:latest --build-arg APPLICATION_VERSION=$(git rev-parse HEAD) .```
`docker-compose up`
The Git sha-1 hash should now be included in the `/status` message.


samiwel:
[Add status pages (3)](https://trello.com/c/R6n1aCJP/250-add-status-pages-3)

PR #51, opened 2018-02-28
378 reorder pages

WickyNilliams:
Add functionality to allow pages to be reordered/moved/inserted at a specific position. 

Dependent on changes here: https://github.com/ONSdigital/eq-author-graphql-schema/pull/22

### How to review 
Run tests, check code is sensible

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,7 +23,7 @@ services:
       - yarn
       - start:dev
   db:
-    image: postgres
+    image: postgres:9.4-alpine
     environment:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=mysecretpassword
--- a/migrations/20180209165440_seed_page_order_column.js
+++ b/migrations/20180209165440_seed_page_order_column.js
@@ -1,0 +1,47 @@
+exports.up = function(knex) {
+  /**
+   * Imagine you have a table like this:
+   * 
+   *   id   SectionId  
+   *  ---- ----------- 
+   *    1           1  
+   *    2           1  
+   *    3           1  
+   *    4           2  
+   *    5           2  
+   *    6           2  
+   * 
+   * 
+   * This query generates a `ROW_NUMBER` (aliased as `position`), partitioned by `SectionId`, and ordered by `id`:
+   * 
+   *   id   SectionId   position  
+   *  ---- ----------- ---------- 
+   *    1           1          1  
+   *    2           1          2  
+   *    3           1          3  
+   *    4           2          1  
+   *    5           2          2  
+   *    6           2          3  
+   * 
+   * 
+   * The `position` value is multiplied by 1000 to gives spaced values for the `order` column.
+   * Each row is then updated with it's corresponding `order` value.
+   * 
+   */
+  return knex.raw(`
+    UPDATE "Pages"
+    SET "order" = p.position
+    FROM (
+      SELECT id, ROW_NUMBER () OVER (
+        PARTITION BY "SectionId"
+        ORDER BY id
+      ) * 1000 as "position"
+      FROM "Pages"
+    ) p
+    WHERE "Pages".id = p.id;
+  `);
+};
+
+exports.down = function(knex) {
+  return knex("Pages").update({ order: 0 });
+};
--- a/repositories/PageRepository.test.js
+++ b/repositories/PageRepository.test.js
@@ -1,0 +1,401 @@
+const db = require("../db");
+const QuestionnaireRepository = require("../repositories/QuestionnaireRepository");
+const SectionRepository = require("../repositories/SectionRepository");
+const PageRepository = require("../repositories/PageRepository");
+const { last, head, map, times } = require("lodash");
+
+const buildQuestionnaire = questionnaire => ({
+  title: "Test questionnaire",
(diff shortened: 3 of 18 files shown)


WickyNilliams:
[Add move page functionality to API (8)](https://trello.com/c/yfOTch1i/378-add-move-page-functionality-to-api-8)

WickyNilliams:
I think it would be good to test this against a copy of the preprod DB. I want to be sure that the migrations work on a large, real data set. To do this purely in a test, I'd need to:

1. run migrations up to the point of master
2. add lots of data
3. run new migrations

But I don't think that knex allows you to run only a subset of migrations, so this isn't possible

WickyNilliams:
There was a change of plan. Going to grab all the data from API manually and write some code to create from mutations.

docker-compose.yml
```diff
@@ -23,7 +23,7 @@ services:
       - yarn
       - start:dev
   db:
-    image: postgres
+    image: postgres:9.4-alpine
```
WickyNilliams: @samiwel this seems to be causing issues for me, when switching from master -> this branch. Not sure what the cause it, but it spits out a bunch of errors then fails to start. Do you get the same?

repositories/PageRepository.test.js
```diff
@@ -0,0 +1,378 @@
...
+          position: 3
+        })
+      );
+
+      const updatedResults = await PageRepository.findAll({
+        sectionId: section.id
+      });
+
+      expect(map(updatedResults, "id")).toEqual(map(pages.reverse(), "id"));
+    });
```
samiwel: Nice work on the tests 👍 Just curious what would happen if a page was deleted, then another page was moved before the first page was un-deleted? would the undeleted page appear at the end i.e. null position?

repositories/QuestionnaireRepository.test.js
```diff
@@ -1,8 +1,4 @@
-const knex = require("../db/index");
-
-const createQuestionnaireTable = require("../migrations/20170620163533_create_questionnaires_table");
-const addSummaryColumn = require("../migrations/20171116131851_add_summary_column");
-const addCreatedByColumn = require("../migrations/20171211100317_add_created_by_to_questionnaires");
+const knex = require("../db");
 const QuestionnaireRepository = require("../repositories/QuestionnaireRepository");
 
 const buildQuestionnaire = (json = {}) => {
```
samiwel: Not necessarily something that needs to be done for this PR. But we have this buildQuestionnaire logic in a few tests. Might be worth refactoring it out into some kind of testUtils file.

tests/setup.js
```diff
@@ -0,0 +1 @@
+jest.setTimeout(10000000);
```
samiwel: What's this for?

migrations/20180209165440_seed_page_order_column.js
```diff
@@ -0,0 +1,19 @@
+exports.up = function(knex) {
+  // TODO: this can probably be converted to knex API rather than `raw`
```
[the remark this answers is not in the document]
WickyNilliams: No, i tried, and it made it _less_ clear. I will remove the TODO. And add some explainer

repositories/PageRepository.test.js
WickyNilliams: `order` values are kept, even for deleted items. So in most cases, the item should return back to the correct position (depending on what moves were made in between delete and undelete). In any case, nothing should be `null`. Could add a specific test for this? It's mostly covered by the existing test `reorders pages correctly even when there are deleted pages in a section`

tests/setup.js
WickyNilliams: Debugging! Jest times out after 10s otherwise, killing the process and making it's impossible to debug without. But maybe it shouldn't be committed?

repositories/PageRepository.test.js
WickyNilliams: here's an illustrative example:

```
# id, order, isDeleted
1, 1000, false
2, 2000, false
3, 3000, false
4, 4000, false 
5, 5000, false

# delete page 4
1, 1000, false
2, 2000, false
3, 3000, false
4, 4000, true 
5, 5000, false

# move page 5
1, 1000, false
2, 2000, false
5, 2500, false
3, 3000, false
4, 4000, true 

# undelete page 4
1, 1000, false
2, 2000, false
5, 2500, false
3, 3000, false
4, 4000, false 
```

tests/setup.js
samiwel: Maybe just add a const explaining the magic number?

repositories/spacedOrderStrategy.js
```diff
@@ -0,0 +1,76 @@
...
+    .andWhere("order", ">", order)
+    .increment("order", SPACING);
+};
+
+const movePage = async (trx, { id, sectionId, position }) => {
+  position = Math.max(0, position);
+
+  let rows = await getAdjacentRows(trx, sectionId, position);
+
+  // position > number of rows, then we need to get max current value
```
nealedj: We just chatted about this function and decided it was a little confusing that the 3 special cases all always run and override `rows`. So the `// inserting at start of list` condition will always take precedence.

It's a little unclear what the intended code paths are at first... or fourth glance 😄

PR #52, opened 2018-03-05
Add otherOption and remove childAnswerId.

samiwel:
This pull request removes childAnswerId from Option as this is an implementation detail of survey runner.

From author's perspective, we just care about whether an Answer (specifically a multiple choice answer) has an Other option or not. In most cases this is defaulted to false, but it is intended that this will be overridden to true for multiple choice answers that have an Other option.

### How to review 
All tests and checks should pass.
API and application should continue to function as normal.

### Dependencies
 - https://github.com/ONSdigital/eq-author-graphql-schema/pull/24


package.json
```diff
@@ -6,22 +6,24 @@
...
     "test": "NODE_ENV=test ./scripts/test.sh",
     "knex": "knex --knexfile config/knexfile.js --cwd .",
     "precommit": "lint-staged"
   },
   "lint-staged": {
-    "*.js": ["prettier --write", "git add"]
+    "*.js": [
+      "prettier --write",
+      "git add"
+    ]
```
WickyNilliams: Maybe we should add a `.prettierignore`?

schema/resolvers.js
```diff
@@ -119,7 +119,42 @@ const Resolvers = {
...
-      ctx.repositories.Option.undelete(args.input.id)
+      ctx.repositories.Option.undelete(args.input.id),
+    createOtherAnswer: async (root, args, ctx) => {
+      const parentAnswer = await ctx.repositories.Answer.get(
+        args.input.parentAnswerId
+      );
+
+      if (!isNil(parentAnswer.otherAnswerId)) {
+        return parentAnswer;
+      }
```
WickyNilliams: Do you think maybe we should throw in this case?

schema/resolvers.js
```diff
@@ -119,7 +119,42 @@ const Resolvers = {
...
+        const otherAnswer = await ctx.repositories.Answer.insert({
+          mandatory: false,
+          type: "TextField"
+        });
+        await ctx.repositories.Answer.update(
+          merge(parentAnswer, { otherAnswerId: otherAnswer.id })
+        );
+      }
+
+      return parentAnswer;
```
WickyNilliams: Why return the parent and not the new answer?

schema/resolvers.js
```diff
@@ -119,7 +119,42 @@ const Resolvers = {
...
+    createOtherAnswer: async (root, args, ctx) => {
+      const parentAnswer = await ctx.repositories.Answer.get(
+        args.input.parentAnswerId
+      );
+
+      if (!isNil(parentAnswer.otherAnswerId)) {
+        return parentAnswer;
+      }
+
+      if (parentAnswer.type === "Radio" || parentAnswer.type === "Checkbox") {
```
WickyNilliams: Again, adding to anything other than these should be an error i think?

WickyNilliams: Also, you could add these as a `where` clause when fetching? Let the DB do the work.

schema/resolvers.js
```diff
@@ -119,7 +119,42 @@ const Resolvers = {
...
+        const otherAnswer = await ctx.repositories.Answer.insert({
+          mandatory: false,
+          type: "TextField"
+        });
+        await ctx.repositories.Answer.update(
+          merge(parentAnswer, { otherAnswerId: otherAnswer.id })
+        );
```
WickyNilliams: Should this be done in a transaction, do you think?

schema/resolvers.js
```diff
@@ -119,7 +119,42 @@ const Resolvers = {
...
+      return parentAnswer;
+    },
+    deleteOtherAnswer: async (_, args, ctx) => {
+      const parentAnswer = await ctx.repositories.Answer.get(
+        args.input.parentAnswerId
+      );
+
+      if (isNil(parentAnswer.otherAnswerId)) {
+        return parentAnswer;
+      }
```
WickyNilliams: As above

schema/resolvers.test.js
```diff
@@ -0,0 +1,161 @@
...
+    let checkboxAnswer = await createNewAnswer(pages[0].id, "Checkbox");
+    expect(checkboxAnswer.otherAnswer).toBeNull();
+
+    checkboxAnswer = await createNewOtherAnswer(checkboxAnswer);
+
+    expect(checkboxAnswer.otherAnswer).not.toBeNull();
+    expect(checkboxAnswer.otherAnswer).toMatchObject({
+      type: "TextField"
+    });
+    done();
```
WickyNilliams: you shouldn't need to call `done()` from any of these. All `async` functions return a promise, and jest automatically waits for promises to resolve

schema/resolvers.js
samiwel: I was originally going to return the child, but decided against it. You can always get the child answer from the parent, but you can't get the parent from the child.

i.e.

schema/resolvers.js
samiwel: I guess there's no harm in throwing here. I was originally going to make it a noop, but I guess the UI should prevent you from trying to add a second otherAnswer and if you invoke it via the API, then it makes more sense to throw I think.

schema/resolvers.test.js
```diff
@@ -0,0 +1,187 @@
...
+    expect(checkboxAnswer.otherAnswer).toBeNull();
+
+    const otherAnswer = await createNewOtherAnswer(checkboxAnswer);
+    expect(otherAnswer).toMatchObject({ type: "TextField" });
+
+    const updatedCheckboxAnswer = await executeQuery(
+      getAnswer,
+      { id: checkboxAnswer.id },
+      ctx
+    ).then(res => res.data.answer);
```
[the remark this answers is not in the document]
WickyNilliams: Also in all other tests

schema/resolvers.js
```diff
@@ -119,7 +125,57 @@ const Resolvers = {
...
+            .parentAnswerId} does not have an "other" answer.`
+        );
+      }
+
+      const otherAnswer = await ctx.repositories.Answer.deleteOtherAnswer(
+        parentAnswer
+      );
+      return ctx.repositories.Answer.update(
+        merge(parentAnswer, { otherAnswerId: null })
+      );
```
WickyNilliams: shouldn't `Answer.deleteOtherAnswer` and the `Answer.update` be part of same transaction?

schema/resolvers.js
```diff
@@ -119,7 +125,57 @@ const Resolvers = {
...
+        throw new Error(
+          "Cannot add a second otherAnswer. Delete the existing one first."
+        );
+      }
+
+      const otherAnswer = await ctx.repositories.Answer.createOtherAnswer(
+        parentAnswer
+      );
+
+      return otherAnswer;
```
WickyNilliams: to keep resolvers simple/clean, can we move as much logic as possible into Answer repository? (both for this and delete other answer)

repositories/AnswerRepository.js
```diff
@@ -4,6 +4,11 @@ const mapFields = require("../utils/mapFields");
 const mapping = { QuestionPageId: "questionPageId" };
 const fromDb = mapFields(mapping);
 const toDb = mapFields(invert(mapping));
+const db = require("../db");
+const {
+  createOtherAnswer,
+  deleteOtherAnswer
+} = require("./strategies/multipleChoiceOtherAnswerStrategy");
 
 module.exports.findAll = function findAll(
```
WickyNilliams: Do we need to exclude "Other" answers here? when selecting regular answers, you want to filter `where({ otherAnswerId : null })`?

samiwel: Yeah good point.

samiwel: I implemented this change, then reverted it, since it broke the resolver for `otherAnswer` property on `MultipleChoiceAnswer` (i.e. prevents you from doing what I mentioned above!).

Filtering out "other" answers should be a concern of resolvers that need to do so.

repositories/strategies/multipleChoiceOtherAnswerStrategy.js
```diff
@@ -0,0 +1,49 @@
...
+const deleteOtherAnswer = async (trx, { id }) => {
+  const existingOtherAnswer = await trx("Answers")
+    .where({ parentAnswerId: id })
+    .first();
+
+  if (isNil(existingOtherAnswer)) {
+    throw new Error(`Answer with id ${id} does not have an "other" answer.`);
+  }
+
+  const otherAnswer = await trx("Answers")
```
WickyNilliams: you can just return here

repositories/strategies/multipleChoiceOtherAnswerStrategy.js
```diff
@@ -0,0 +1,49 @@
...
+    .where({ parentAnswerId: id })
+    .first();
+
+  if (!isNil(existingOtherAnswer)) {
+    throw new Error(
+      "Cannot add a second 'other' Answer. Delete the existing one first."
+    );
+  }
+
+  const otherAnswer = await trx("Answers")
```
WickyNilliams: you can return here

schema/resolvers.test.js
```diff
@@ -0,0 +1,165 @@
...
+
+  it("should not create a new otherAnswer if one already exists", async () => {
+    const parentAnswer = await createNewAnswer(pages[0].id, "Checkbox");
+    const otherAnswer = await createNewOtherAnswer(parentAnswer);
+
+    expect(createNewOtherAnswer(parentAnswer)).resolves.toBeNull();
+
+    const updatedParent = await refreshAnswerDetails(parentAnswer.id);
+    expect(updatedParent.otherAnswer).toMatchObject(otherAnswer);
+  });
```
WickyNilliams: Can you add a test for "should exclude otherAnswer from regular answers"?

PR #64, opened 2018-04-06
Implement API changes in order to support add "other" option and text field answer.

samiwel:
This PR is part of the change required in order implement the "add other option" feature.

This PR implements the necessary logic to resolve the "other" Option and BasicAnswer for a MultipleChoiceAnswer.

The strategy for creating the other Answer has also been updated so that the "other" Option is created and deleted along with the answer.

### How to review 
All tests and checks should pass. Code review required.

### Next steps
Once this PR is merged, there will be changes required in Author and Publisher.

### DIFF
--- a/repositories/OptionRepository.js
+++ b/repositories/OptionRepository.js
@@ -11,7 +11,7 @@ module.exports.findAll = function findAll(
   direction = "asc"
 ) {
   return Option.findAll()
-    .where({ isDeleted: false })
+    .where({ isDeleted: false, otherAnswerId: null })
     .where(where)
     .orderBy(orderBy, direction)
     .then(map(fromDb));
@@ -72,4 +72,15 @@ module.exports.undelete = function(id) {
   return Option.update(id, { isDeleted: false })
     .then(head)
     .then(fromDb);
+};
+
+module.exports.getOtherOption = function(
+  answerId,
+  orderBy = "created_at",
+  direction = "asc"
+) {
+  return Option.findAll()
+    .where({ isDeleted: false, otherAnswerId: answerId })
+    .orderBy(orderBy, direction)
+    .first();
 };
--- a/repositories/strategies/multipleChoiceOtherAnswerStrategy.js
+++ b/repositories/strategies/multipleChoiceOtherAnswerStrategy.js
@@ -1,9 +1,48 @@
 const { head, isNil } = require("lodash/fp");
 
-const createOtherAnswer = async (trx, { id }) => {
-  const existingOtherAnswer = await trx("Answers")
-    .where({ parentAnswerId: id })
+const findOtherAnswer = async (trx, parentAnswerId) =>
+  trx("Answers")
+    .where({ parentAnswerId, isDeleted: false })
     .first();
+
+const createAnswer = async (trx, parentAnswerId, type) =>
+  trx("Answers")
+    .insert({
+      mandatory: false,
+      type,
+      parentAnswerId
+    })
+    .returning("*")
+    .then(head);
+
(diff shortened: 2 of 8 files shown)


repositories/OptionRepository.js
```diff
@@ -71,3 +71,16 @@ module.exports.remove = function remove(id) {
...
 };
+
+module.exports.getOtherOption = function(
+  id,
+  where = {},
+  orderBy = "created_at",
+  direction = "asc"
+) {
+  return Option.findAll()
+    .where({ isDeleted: false, otherAnswerId: id })
```
SamGodwin2: Why don't we just make the default value = `{ isDeleted: false, otherAnswerId: id }` then if we want to use a different query for any reason we can.

repositories/strategies/multipleChoiceOtherAnswerStrategy.js
```diff
@@ -1,42 +1,69 @@
 const { head, isNil } = require("lodash/fp");
 
-const createOtherAnswer = async (trx, { id }) => {
-  const existingOtherAnswer = await trx("Answers")
-    .where({ parentAnswerId: id })
+const findOtherAnswer = async (trx, parentAnswerId) =>
+  trx("Answers")
+    .where({ parentAnswerId, isDeleted: false })
     .first();
```
SamGodwin2: Can we remove the first? Only ever going to be one option that fulfills both of those conditions?

samiwel: The call to `first` resolves the promise and returns the first result. Without that we'd need to await the result of a select and then get the first element of that array ourselves. I think using first is a cleaner implementation in this case.

repositories/OptionRepository.js
samiwel: After thinking about it some more, I'm not sure we should do this. I can't think of a case where we'd ever want to retrieve the deleted other option, and we'd always want to get the Other entry for a specific answer Id.

Equally, I don't think we'd ever want to specialise this query any further. So it might make more sense to remove the second where clause altogether.

schema/resolvers.js
```diff
@@ -187,8 +187,18 @@ const Resolvers = {
...
+    other: async ({ id }, args, ctx) => {
+      try {
+        const answer = await ctx.repositories.Answer.getOtherAnswer(id);
+        const option = await ctx.repositories.Option.getOtherOption(answer.id);
+        return {
+          answer,
+          option
+        };
+      } catch (e) {
+        return null;
```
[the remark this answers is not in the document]
samiwel: Could do it that way but it led to more return points.
 
e.g.
```
if (isNil(answer)) return null;
const option = await getOption(answer.id);
if (isNil(option)) return null;
return {
	answer,
    option
}
```

I thought the try catch was more readable, but happy to change it if you disagree?

samiwel:
[Add/edit 'other' answer (8)](https://trello.com/c/3i8vZj7b/70-add-edit-other-answer-8)

repositories/strategies/multipleChoiceOtherAnswerStrategy.js
```diff
@@ -1,42 +1,69 @@
...
+    .then(head);
+
+const deleteAnswer = async (trx, { id }) =>
+  trx("Answers")
+    .where("id", id)
+    .update({
+      isDeleted: true
+    })
+    .returning("*")
+    .then(head);
```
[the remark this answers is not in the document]
samiwel: Yeah there's some duplication here. There's not an easy way to re-use the existing code within the scope of the transaction. Will create a tech debt card for that refactoring work.

PR #68, opened 2018-04-18
343 move page tweaks

WickyNilliams:
Adds ability to get `position` for a single page in isolation. Extract some common code for reuse.

### How to review 
Code looks good, tests pass

### DIFF
--- a/repositories/PageRepository.js
+++ b/repositories/PageRepository.js
@@ -5,7 +5,9 @@ const mapFields = require("../utils/mapFields");
 const db = require("../db");
 const {
   movePage,
-  getNextOrderValue
+  getNextOrderValue,
+  calculatedPositionCol,
+  getPosition
 } = require("./strategies/spacedOrderStrategy");
 
 const mapping = { SectionId: "sectionId" };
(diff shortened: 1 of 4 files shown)


WickyNilliams:
[Re-order(move) pages (UI) (5)](https://trello.com/c/ujl21Rsf/343-re-ordermove-pages-ui-5)

repositories/PageRepository.js
```diff
@@ -86,5 +86,13 @@ module.exports.undelete = function(id) {
...
 
 module.exports.move = ({ id, sectionId, position }) => {
-  return db.transaction(trx => movePage(trx, { id, sectionId, position }));
+  return db.transaction(trx =>
+    movePage(trx, { id, sectionId, position }).then(fromDb)
+  );
+};
+
+module.exports.getPosition = ({ id }) => {
+  return Page.findById(id).then(({ SectionId }) =>
```
hamishtaplin: what's the reason for capitalising `SectionId` here?

WickyNilliams: Because I made the foolish design choice right at the start of this project to use PascalCase for foreign keys in the DB, but expose them as camelCase to clients.

I could add a `then(fromDb)` here to avoid it. But felt a bit much since it's only a temporary value, and this method returns a single integer (not a row or set of rows)

repositories/PageRepository.test.js
```diff
@@ -357,8 +357,8 @@ describe("PagesRepository", () => {
         buildPage({ sectionId: section.id, position: 1 })
       );
 
-      // by moving 10 times the `order` values will converge - (1000 / 2^10) < 1
-      const plan = times(10, i => (i % 2 === 0 ? page1 : page2));
+      // by moving more than 10 times the `order` values will converge, since (1000 / 2^12) < 1
+      const plan = times(12, i => (i % 2 === 0 ? page1 : page2));
```
samiwel: What was the reason for changing this? Moving 10 times and the values would also converge, right? since 2^10 = 1024 and 1000/1024 would still be < 1?

PR #70, opened 2018-04-26
consistently map fields from db

WickyNilliams:
Use `fromDb` helper to map fields name from DB naming convention to Schema naming convention. This ensures that we can always select fields via graphql query

### How to review 
Tests pass, changes look good


WickyNilliams:
[BUG: undeleting questionnaire or pages inserts at incorrect position](https://trello.com/c/8u7Dqmnm/424-bug-undeleting-questionnaire-or-pages-inserts-at-incorrect-position)

PR #73, opened 2018-05-02
Fixes null description passed through for `other` answer.

samiwel:
There is an issue where null values are being returned for other answer description.

This change to the API should ensure that other answers have an empty string
description by default.

This is only an issue because the other answer is a hidden field. So users cannot
provide a value.

### How to review 
 - Code changes look good.
 - All tests pass.


samiwel:
[Add/edit 'other' answer (8)](https://trello.com/c/3i8vZj7b/70-add-edit-other-answer-8)

PR #74, opened 2018-05-04
Improvements to move page feature

WickyNilliams:
* Simplifies some awkward logic around computing the `position` value for pages. Possibly needs some indexes or something added. Need to investigate further. But for now the code works
* Simplifies logic around calculating `order` values when moving pages
* Reduces number of round trips to DB, at cost of bigger payload from DB
* Improves performance of inserts - previously would insert at end of list, _then_ perform a move operation to place at correct position. Now inserts directly at correct position

### How to review 
Run tests. Run e2e tests. Check code is sensible

### DIFF
--- a/repositories/PageRepository.js
+++ b/repositories/PageRepository.js
@@ -4,8 +4,7 @@ const QuestionPageRepository = require("./QuestionPageRepository");
 const mapFields = require("../utils/mapFields");
 const db = require("../db");
 const {
-  movePage,
-  getNextOrderValue
+  getOrUpdateOrderForInsert
 } = require("./strategies/spacedOrderStrategy");
 
 const mapping = { SectionId: "sectionId" };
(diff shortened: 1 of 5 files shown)


repositories/PageRepository.js
```diff
@@ -68,33 +62,42 @@ module.exports.insert = function insert(args) {
...
+Object.assign(module.exports, {
+  findAll,
+  get: getById,
+  insert,
+  update,
+  remove,
+  undelete,
+  move,
+  getPosition
+});
```
samiwel: Nice tidy up 👍

WickyNilliams: It's a clever trick I saw here: https://twitter.com/getify/status/991427888359596032.

It saves on repetitious `module.export.whatever`, and prevents circular dependencies from causing issues.

WickyNilliams:
[Consider using View table to simplify data access logic around reordering](https://trello.com/c/HePJ415f/423-consider-using-view-table-to-simplify-data-access-logic-around-reordering)

repositories/strategies/spacedOrderStrategy.js
```diff
@@ -1,104 +1,54 @@
...
+
+  return isNil(id) ? query : query.where("id", "!=", id);
 };
 
-const movePage = async (trx, { id, sectionId, position }) => {
-  position = Math.max(0, position);
+const getMaxOrder = flow(last, getOr(0, "order"));
 
-  let rows = await getAdjacentRows(trx, id, sectionId, position);
+const makeSpaceForInsert = async (trx, sectionId, id, position) => {
```
WickyNilliams: Not sure what to call this function. This doesn't seem like a great name, as it doesn't always "make space". Sometimes it's just doing a calculation, and others it's updating rows in the DB.

Anyone have any thoughts on this?

SamGodwin2: Could (get/find)OrderForInsert() work?

PR #77, opened 2018-05-09
Implement routing on api

SamGodwin2:
This PR adds necessary resolvers for the newly introduced GraphQL schema types in order to support basic routing functionality within Author.

Changes include:
 - New migration scripts to create the new database tables.
 - Resolver functions for each of the new GraphQL types for basic routing.
 - Database and transaction logic (Repository, queries etc.)
 - New integration and unit tests covering the core routing logic.
 - New field arguments to support bringing back section and page titles as plaintext.

### How to review 
All checks and tests should pass.
Code review required.

### Dependencies
 - https://github.com/ONSdigital/eq-author-graphql-schema/pull/37

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:8-alpine
 EXPOSE 4000
 ENV PORT=4000
 WORKDIR /app
@@ -11,4 +11,3 @@ ENTRYPOINT ["yarn", "start"]
 
 COPY . /app
 RUN yarn install
-RUN yarn upgrade eq-author-graphql-schema
--- a/db/Routing.js
+++ b/db/Routing.js
@@ -1,0 +1,79 @@
+const db = require("./");
+const { parseInt } = require("lodash");
+const RoutingRuleSets = "Routing_RuleSets";
+const RoutingRules = "Routing_Rules";
+const RoutingConditions = "Routing_Conditions";
+const RoutingConditionValues = "Routing_ConditionValues";
+const RoutingDestinations = "Routing_Destinations";
+
+const table = (tableName, knex = db) => knex(tableName);
+const select = (tableName, knex = db) => table(tableName, knex).select();
+const selectById = (tableName, id, knex = db) =>
+  select(tableName, knex)
+    .where("id", parseInt(id))
+    .first();
+const insert = (tableName, record, knex = db) =>
+  table(tableName, knex)
+    .insert(record)
+    .returning("*");
+const update = (tableName, id, record, knex = db) =>
+  table(tableName, knex)
+    .where("id", parseInt(id))
+    .update(record)
+    .returning("*");
+const deleteById = (tableName, id, knex = db) =>
+  table(tableName, knex)
+    .where({ id })
+    .del()
+    .returning("*");
+
+const findAllRoutingRuleSets = () => select(RoutingRuleSets);
+const findAllRoutingRules = () => select(RoutingRules);
+const findAllRoutingConditions = () => select(RoutingConditions);
+const findAllRoutingConditionValues = () => select(RoutingConditionValues);
+
+const findRoutingRuleSetsById = id => selectById(RoutingRuleSets, id);
+const findRoutingRulesById = id => selectById(RoutingRules, id);
+const findRoutingConditionValuesById = id =>
+  selectById(RoutingConditionValues, id);
+const findRoutingDestinationById = id => selectById(RoutingDestinations, id);
+
+const createRoutingRuleSet = values => insert(RoutingRuleSets, values);
+const createRoutingRule = values => insert(RoutingRules, values);
+const createRoutingCondition = values => insert(RoutingConditions, values);
+const createRoutingConditionValue = values =>
+  insert(RoutingConditionValues, values);
+
+const updateRoutingRuleSet = (id, values) =>
+  update(RoutingRuleSets, id, values);
+const updateRoutingRule = (id, values) => update(RoutingRules, id, values);
+const updateRoutingCondition = (id, values) =>
+  update(RoutingConditions, id, values);
+const updateRoutingConditionValue = (id, values) =>
+  update(RoutingConditionValues, id, values);
+
+const updateRoutingDestination = (id, values) =>
+  update(RoutingDestinations, id, values);
+
+const deleteRoutingCondition = id => deleteById(RoutingConditions, id);
+
+Object.assign(module.exports, {
+  findAllRoutingRuleSets,
+  findAllRoutingRules,
+  findAllRoutingConditions,
+  findAllRoutingConditionValues,
+  findRoutingRuleSetsById,
+  findRoutingRulesById,
+  findRoutingConditionValuesById,
+  createRoutingRuleSet,
+  createRoutingRule,
+  createRoutingCondition,
+  createRoutingConditionValue,
+  updateRoutingRuleSet,
+  updateRoutingRule,
+  updateRoutingCondition,
+  updateRoutingConditionValue,
+  deleteRoutingCondition,
+  findRoutingDestinationById,
+  updateRoutingDestination
+});
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       - db
     volumes:
       - .:/app
-      - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
+      # - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
     ports:
       - 4000:4000
       - 5858:5858 # open port for debugging
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       - db
     volumes:
       - .:/app
-      # - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
+      - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
     ports:
       - 4000:4000
       - 5858:5858 # open port for debugging
--- a/migrations/20180430155015_creating_routing_databases.js
+++ b/migrations/20180430155015_creating_routing_databases.js
@@ -1,0 +1,140 @@
+const createRoutingRuleSetTable = async knex => {
+  return knex.schema.createTable("Routing_RuleSets", function(table) {
+    table.increments();
+
+    table
+      .integer("QuestionPageId")
+      .unsigned()
+      .references("id")
+      .inTable("Pages")
+      .onDelete("CASCADE");
+
+    table
+      .integer("RoutingDestinationId")
+      .unsigned()
+      .references("id")
+      .inTable("Routing_Destinations")
+      .onDelete("CASCADE");
+
+    table
+      .boolean("isDeleted")
+      .notNull()
+      .defaultTo(false);
+  });
+};
+
+const createRoutingRuleTable = async knex => {
+  return knex.schema.createTable("Routing_Rules", function(table) {
+    table.increments();
+
+    table.enum("operation", ["And", "Or"]).notNullable();
+
+    table
+      .integer("RoutingRuleSetId")
+      .unsigned()
+      .references("id")
+      .inTable("Routing_RuleSets")
+      .onDelete("CASCADE");
+
+    table
+      .integer("RoutingDestinationId")
+      .unsigned()
+      .references("id")
+      .inTable("Routing_Destinations")
+      .onDelete("CASCADE");
+
+    table
+      .boolean("isDeleted")
+      .notNull()
+      .defaultTo(false);
+  });
+};
+
+const createRoutingConditionTable = async knex => {
+  return knex.schema.createTable("Routing_Conditions", function(table) {
+    table.increments();
+
+    table.enum("comparator", ["Equal", "NotEqual"]).notNullable();
+
+    table
+      .integer("RoutingRuleId")
+      .unsigned()
+      .references("id")
+      .inTable("Routing_Rules")
+      .onDelete("CASCADE");
+
+    table
+      .integer("QuestionPageId")
+      .unsigned()
+      .references("id")
+      .inTable("Pages")
+      .onDelete("CASCADE");
+
+    table
+      .integer("AnswerId")
+      .unsigned()
+      .references("id")
+      .inTable("Answers")
+      .onDelete("CASCADE");
+  });
+};
+
+const createRoutingConditionValuesTable = async knex => {
+  return knex.schema.createTable("Routing_ConditionValues", function(table) {
+    table.increments();
+
+    table
+      .integer("OptionId")
+      .unsigned()
+      .references("id")
+      .inTable("Options")
+      .onDelete("CASCADE");
+
+    table
+      .integer("ConditionId")
+      .unsigned()
+      .references("id")
+      .inTable("Routing_Conditions")
+      .onDelete("CASCADE");
+  });
+};
+
+const createRoutingDestinationsTable = async knex => {
+  return knex.schema.createTable("Routing_Destinations", table => {
+    table.increments();
+
+    table
+      .enum("logicalDestination", ["NextPage", "EndOfQuestionnaire"])
+      .defaultTo("NextPage");
+
+    table
+      .integer("PageId")
+      .unsigned()
+      .references("id")
+      .inTable("Pages")
+      .onDelete("CASCADE");
+
+    table
+      .integer("SectionId")
+      .unsigned()
+      .references("id")
+      .inTable("Sections")
+      .onDelete("CASCADE");
+  });
+};
+
+exports.up = async function(knex) {
+  await createRoutingDestinationsTable(knex);
+  await createRoutingRuleSetTable(knex);
+  await createRoutingRuleTable(knex);
+  await createRoutingConditionTable(knex);
+  await createRoutingConditionValuesTable(knex);
+};
+
+exports.down = async function(knex) {
+  await knex.schema.dropTable("Routing_ConditionValues");
+  await knex.schema.dropTable("Routing_Conditions");
+  await knex.schema.dropTable("Routing_Rules");
+  await knex.schema.dropTable("Routing_RuleSets");
+  await knex.schema.dropTable("Routing_Destinations");
+};
--- a/repositories/RoutingRepository.js
+++ b/repositories/RoutingRepository.js
@@ -1,0 +1,274 @@
+const { head, invert, map } = require("lodash/fp");
+const { get, isNil, parseInt } = require("lodash");
+const db = require("../db");
+const mapFields = require("../utils/mapFields");
+
+const {
+  updateRoutingConditionStrategy,
+  updateRoutingConditionValueStrategy,
+  getAvailableRoutingDestinations,
+  checkRoutingDestinations,
+  createRoutingRuleSetStrategy,
+  createRoutingRuleStrategy,
+  createRoutingConditionStrategy,
+  resolveRoutingDestinationStrategy
+} = require("./strategies/routingStrategy");
+
+const Routing = require("../db/Routing");
+const PageRepository = require("./PageRepository");
+const SectionRepository = require("./SectionRepository");
+
+const mapping = {
+  QuestionPageId: "questionPageId",
+  RoutingRuleSetId: "routingRuleSetId",
+  RoutingRuleId: "routingRuleId",
+  AnswerId: "answerId",
+  OptionId: "optionId",
+  ConditionId: "conditionId",
+  PageId: "pageId",
+  SectionId: "sectionId",
+  RoutingDestinationId: "routingDestinationId"
+};
+const fromDb = mapFields(mapping);
+const toDb = mapFields(invert(mapping));
+
+const getRoutingDestinations = async pageId => {
+  const logicalDestinations = [
+    {
+      logicalDestination: "NextPage"
+    },
+    {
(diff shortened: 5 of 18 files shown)


SamGodwin2:
[Implement routing on API (8)](https://trello.com/c/SWcislz8/361-implement-routing-on-api-8)

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+const updateConditionsTable = async (trx, { id, answerId }) =>
+  trx("Routing_Conditions")
+    .where({ id })
+    .update({
+      AnswerId: answerId
+    })
+    .returning("*")
+    .then(head);
+
+const updateValuesTable = async (trx, { id }) =>
```
samiwel: Rename `id` to `conditionId`.

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+  trx("Sections")
+    .select("QuestionnaireId", "id")
+    .where({ id: SectionId })
+    .then(head);
+
+const getDestinationsInSection = async (trx, { SectionId, order }) =>
+  trx("Pages")
+    .select("id")
+    .where({ SectionId })
+    .where("order", ">", order);
```
samiwel: Won't this select deleted pages as well?

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+      .where({ OptionId: optionId })
+      .del()
+      .returning("*")
+      .then(head);
+  }
+};
+
+module.exports.getAvailableRoutingDestinations = async function getAvailableRoutingDestinations(
+  trx,
+  id
```
samiwel: Rename to `pageId`?

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+) {
+  const sectionId = await getSectionId(trx, id);
```
samiwel: I think this should be `page` and `getPage()` it wasn't immediately obvious that getSectionId queries the pages table and returns order in addition to the page's sectionId.

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+  const questionnaireId = await getQuestionnaireId(trx, sectionId);
```
samiwel: `getQuestionnnaireId` is really getSection? then you can say something like

const questionnaireId = getSection().questionnaireId?

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+  const destinationsInSection = await getDestinationsInSection(trx, sectionId);
+  const sectionsAheadOfPage = await getSectionsAheadOfPage(
+    trx,
+    questionnaireId,
+    sectionId
+  );
```
samiwel: I don't think this function takes into consideration deleted sections.

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+  { QuestionnaireId },
+  { SectionId }
+) =>
+  trx("Sections")
+    .select("id")
+    .where({ QuestionnaireId })
+    .where("id", ">", SectionId);
+
+const getDestinationsOutsideSection = async (trx, sectionIds) => {
+  const idArray = sectionIds.map(sectionId => sectionId.id);
```
samiwel: `const sectionIds = sections.map(section => section.id);`

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+  return trx("Pages")
+    .select("id")
+    .from(function() {
+      this.select("SectionId")
+        .from("Pages")
```
samiwel: Again we probably don't care about deleted pages here?

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+        "firstPages.min",
+        "=",
+        "Pages.order"
+      );
+    });
+};
+
+module.exports.updateConditions = async function updateConditions(
+  trx,
+  { id, answerId }
```
samiwel: `{ conditionId, answerId }`

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+) {
+  const prevAnswerId = await getConditionAnswerId(trx, { id });
+  if (parseInt(answerId) === prevAnswerId.AnswerId) {
+    return;
```
samiwel: What's the scenario where this would be called? Would we ever fire an update from the UI if the answer doesn't change?

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+  } else {
+    await updateValuesTable(trx, { id });
```
samiwel: Rename to `deleteRoutingConditionValues` to convey intent?

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,127 @@
...
+  );
+  const destinationsOutsideSection = await getDestinationsOutsideSection(
+    trx,
+    sectionsAheadOfPage
+  );
+  const destinations = concat(
+    destinationsInSection,
+    destinationsOutsideSection
+  );
+  return destinations.map(destination => destination.id);
```
samiwel: I'd be tempted to return the destinations here. Let whatever is calling this get the Ids?

repositories/strategies/routingStrategy.js
SamGodwin2: I wasn't sure, does seem unlikely that there would be an update if nothing changes. Just felt a little warranted as we're doing hard deletes here?

repositories/RoutingRepository.js
```diff
@@ -0,0 +1,172 @@
...
+const mapping = {
+  QuestionPageId: "questionPageId",
+  ElseDestination: "elseDestination",
+  RoutingRuleSetId: "routingRuleSetId",
+  RuleDestination: "ruleDestination",
+  RoutingRuleId: "routingRuleId",
+  AnswerId: "answerId",
+  OptionId: "optionId",
+  ConditionId: "conditionId"
+};
```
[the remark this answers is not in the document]
SamGodwin2: I wouldn't mind this remembering which fields were capitalised was a pain, but wouldn't that come away from the existing pattern where fk column names are capitalised?

repositories/RoutingRepository.js
```diff
@@ -0,0 +1,171 @@
...
+
+module.exports.findAllRoutingConditionValues = function findAllRoutingConditionValues(
+  where = {}
+) {
+  return Routing.findAllRoutingConditionValues()
+    .where(where)
+    .map(result => result.OptionId);
+};
+
+module.exports.insertRoutingRuleSet = function insertRoutingRuleSet({
```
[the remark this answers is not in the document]
SamGodwin2: Can do but I'm not sure it would make the implementation much cleaner as it would just change the order of the names 

i.e. Routing.findAllRoutingRules() => RoutingRules.findAll()

and it would also just spread the functions over a number of files. If we're sure that implementation is better I can refactor though.

repositories/RoutingRepository.js
```diff
@@ -0,0 +1,169 @@
...
+    .then(map(fromDb));
+};
+
+module.exports.findAllRoutingConditionValues = function findAllRoutingConditionValues(
+  where = {}
+) {
+  return Routing.findAllRoutingConditionValues().where(where);
+};
+
+module.exports.insertRoutingRuleSet = function insertRoutingRuleSet({
```
samiwel: Something that just occurred to me. We probably need some kind of check here to guard against being able to supply an else destination that is not in the available routing destinations. I know we'll limit the options available to the author in the UI, but we should probably check for this at the API level as well.

e.g.
```js
if (elseDestination != getAvailableRoutingDestinations()) {
	throw new Error(`Unable to route from ${questionPageId} to ${elseDestination}`);
}

Comment (samiwel) [reply repositories/RoutingRepository.js]: Also probably need some similar guard for the `goto` field of RoutingRule.

File: repositories/RoutingRepository.js [hunk 16]
```diff
@@ -0,0 +1,169 @@
...
+
+module.exports.removeRoutingCondition = function removeRoutingCondition(id) {
+  return Routing.deleteRoutingCondition(id).then(head);
+};
+
+module.exports.undeleteRoutingRule = function undeleteRoutingRule(id) {
+  return Routing.updateRoutingRule(id, { isDeleted: false })
+    .then(head)
+    .then(fromDb);
+};
```
samiwel: This file would benefit from Nick's Object.assign exports clean-up.

db/Routing.js
```diff
@@ -0,0 +1,124 @@
...
+    .update(updates)
+    .returning("*");
+};
+
+module.exports.deleteRoutingCondition = function deleteRoutingCondition(id) {
+  return RoutingCondition()
+    .where({ id })
+    .returning("*")
+    .del();
+};
```
samiwel: This file would benefit from Nick's Object.assign exports clean-up.

migrations/20180430155015_creating_routing_databases.js
```diff
@@ -0,0 +1,128 @@
...
+      .references("id")
+      .inTable("Sections")
+      .onDelete("CASCADE");
+
+    table
+      .integer("PageDestination")
+      .unsigned()
+      .references("id")
+      .inTable("Pages")
+      .onDelete("CASCADE");
```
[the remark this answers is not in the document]
samiwel: In this case, each of the foreign keys has a delete cascade. So the routing ruleset will be deleted whenever the question page that owns it is deleted, or when the section or page being referred to in the rule's ELSE route is deleted.

In practice this won't ever be fired due to soft deletes being used in the application, but I can see that being able to clear out your data quickly during development might be appealing.

repositories/RoutingRepository.js
```diff
@@ -0,0 +1,229 @@
...
+
+async function createRoutingRule(createRoutingRuleInput) {
+  const sectionDestination = get(
+    createRoutingRuleInput,
+    "goto.sectionId",
+    null
+  );
+  const pageDestination = get(createRoutingRuleInput, "goto.pageId", null);
+
+  if (some([sectionDestination, pageDestination], !isNil)) {
```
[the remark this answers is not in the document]
WickyNilliams: Yeah sorry my example was off. And I don't think I explained very well!

This code does not work as expected:

```js
some([sectionDestination, pageDestination], !isNil);
```

You're using a boolean operator on a function, so you're passing `false` as the predicate function. What you likely meant was `negate(isNil)`.

In any case, I think expanding the if statement is better. It feels weird to create an array to check two values. If you like you can refactor the boolean expression`!A || !B` to `!(A && B)`

repositories/RoutingRepository.js
```diff
@@ -0,0 +1,229 @@
...
+}
+
+function undeleteRoutingRule(id) {
+  return Routing.updateRoutingRule(id, { isDeleted: false })
+    .then(head)
+    .then(fromDb);
+}
+
+function getRoutingDestinations(pageId) {
+  return db.transaction(trx => getAvailableRoutingDestinations(trx, pageId));
```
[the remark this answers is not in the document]
samiwel: The results of `getAvailableRoutingDestinations` are returned having already been mapped over with `fromDb`.

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,287 @@
...
+      .then(fromDb);
+  } else {
+    return table
+      .where(toDb({ conditionId, optionId }))
+      .del()
+      .returning("*")
+      .then(head)
+      .then(fromDb);
+  }
+};
```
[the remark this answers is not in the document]
samiwel: Yeah way better.

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,287 @@
...
+
+async function getAvailableRoutingDestinations(trx, pageId) {
+  const page = await getPageById(trx, pageId);
+  const section = await getSectionById(trx, page.sectionId);
+  const pageDestinations = await getPageDestinations(trx, page);
+  const sectionDestinations = await getSectionDestinations(trx, section);
```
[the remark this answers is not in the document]
samiwel: Good shout, ordering is performed in the front-end, so this is a good optimisation.

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,287 @@
...
+    routingRuleId: routingRule.id,
+    questionPageId: routingRuleSet.questionPageId
+  });
+
+  return routingRule;
+}
+
+async function createRoutingRuleSetStrategy(trx, questionPageId) {
+  const existingRuleSet = await findRoutingRuleSet(trx, questionPageId);
+  if (!isNil(existingRuleSet)) {
```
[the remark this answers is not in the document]
samiwel: @WickyNilliams How do you mean?

schema/resolvers.js
```diff
@@ -176,7 +209,83 @@ const Resolvers = {
...
+      });
+    },
+    goto: ({ sectionDestination, pageDestination }, args, ctx) => {
+      if (isNil(sectionDestination)) {
+        return null;
+      } else if (isNil(pageDestination)) {
+        return ctx.repositories.Section.get(sectionDestination);
+      } else {
+        return ctx.repositories.Page.get(pageDestination);
+      }
```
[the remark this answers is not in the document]
samiwel: Yeah its a bit of a smell. For now, I can extract the logic into one place. We can revisit RoutingDestination again. Semantically a routing destination is either a Section or a Page, and can optionally not be set, so the logic is correct.

There was some debate about whether RoutingDestination should be a type in its own right. We opted against it because its really nothing more than a join table between routingCondition, routingRuleSet, section and questionPage. We can always re-evaluate if it makes sense to migrate to this in future.

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,287 @@
...
+) {
+  const partitionedDestinations = partition(
+    availableRoutingDestinations,
+    "sectionId"
+  );
+  const questionPages = first(partitionedDestinations);
+  const sections = last(partitionedDestinations);
+
+  if (isNil(pageDestination)) {
+    if (!find(sections, { id: parseInt(sectionDestination) })) {
```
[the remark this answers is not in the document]
samiwel: Actually, includes won't work here since it compares values and doesn't perform identity checking. https://stackoverflow.com/a/25171174. Have reverted back to `find` since it reads better than using `some`.

repositories/strategies/routingStrategy.js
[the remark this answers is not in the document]
WickyNilliams: i mean can you use a unique constraint in the DB to ensure that you can't have more than one rule set? since this code checks for an existing one then throws, can we have the DB enforce it for us instead?

repositories/strategies/routingStrategy.js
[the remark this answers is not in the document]
WickyNilliams: Ah good catch. Didn't realise includes worked like that! Yeah, `find` is the best choice in that case

repositories/strategies/routingStrategy.js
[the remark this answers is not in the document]
samiwel: Oh I see, yeah that makes sense. The only reason I can see for not doing that is that RoutingRuleSets have an `isDeleted` column meaning they are intended for soft deletion.

In that case it's plausible you might have more than one RoutingRuleSet with the same page Id. In practice, we have no way to delete the routing rule set at the moment. Depends if deleting rule sets (i.e. removing all routing rules) becomes a thing?

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,287 @@
...
+};
+
+const toggleConditionValueStrategy = async (
+  trx,
+  { conditionId, optionId, checked }
+) => {
+  const table = trx("Routing_ConditionValues");
+  const args = toDb({ conditionId, optionId });
+  const existing = await table.where(args);
+  const shouldInsert = checked || isEmpty(existing);
```
SamGodwin2: Should probably throw an error here if `!isEmpty(existing)&&checked` then the query can just check checked to see if its an insert or delete query.

schema/resolvers.test.js
```diff
@@ -225,4 +404,343 @@ describe("resolvers", () => {
...
+        });
+
+        const updated = await getRoutingDataForPage(firstPage);
+        expect(
+          get(updated, "routingRuleSet.routingRules[0].conditions[0]")
+        ).toMatchObject({
+          questionPage: {
+            id: newPage.id
+          }
+        });
```
SamGodwin2: Possibly want a test here somewhere to make sure two condition values with the same args can't be inserted?

db/Routing.js
```diff
@@ -0,0 +1,72 @@
...
+    .where("id", parseInt(id))
+    .update(record)
+    .returning("*");
+const deleteById = (tableName, id, knex = db) =>
+  table(tableName, knex)
+    .where({ id })
+    .del()
+    .returning("*");
+
+const findAllRoutingRuleSets = () => select(RoutingRuleSets);
```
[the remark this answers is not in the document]
samiwel: I've always wanted to find a good use for `curry`ing. Now I have one!

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:8-alpine
```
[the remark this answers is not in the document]
samiwel: I think the other task is more-generally across all survey runner and author repos. But this is a step in that direction.

db/Routing.js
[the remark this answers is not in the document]
samiwel: As discussed will leave as-is for now as we have a separate card to look at refactoring some of the DB logic, which would be a good opportunity to review our approach.

repositories/strategies/routingStrategy.js
```diff
@@ -183,37 +189,47 @@ const toggleConditionValueStrategy = async (
...
-    if (!find(sections, { id: parseInt(sectionDestination) })) {
+  if (!isNil(logicalDestination)) {
+    if (
+      !find(logicalDestinations, {
+        logicalDestination: logicalDestination.destinationType
+      })
+    ) {
       throw new Error(
-        `Unable to route from this question to section ${sectionDestination}`
+        `Unable to route from this question ${logicalDestination.destinationType}`
```
SamGodwin2: Could you do this on with one if statement instead by using 
`if(includes(logicalDestinations, logicalDestination.destinationType))`

schema/resolvers.test.js
```diff
@@ -8,11 +8,29 @@ const {
...
+  deleteRoutingRuleSet,
+  createRoutingRule,
+  deleteRoutingRule,
+  createRoutingCondition,
+  deleteRoutingCondition
 } = require("../tests/utils/graphql");
 
 const ctx = { repositories };
 
+jest.setTimeout(999999999);
```
SamGodwin2: Meant to be here?

repositories/strategies/routingStrategy.js
```diff
@@ -0,0 +1,322 @@
...
+  if (!isNil(logicalDestination)) {
+    if (
+      !find(logicalDestinations, {
+        logicalDestination: logicalDestination.destinationType
+      })
+    ) {
+      throw new Error(
+        `Unable to route from this question ${logicalDestination.destinationType}`
+      );
+    }
```
SamGodwin2: Could you do this in one line using `if(includes(logicalDestinations, logicalDestination.destinationType))`

SamGodwin2: Could you do this in one line using if(includes(logicalDestinations, logicalDestination.destinationType))

docker-compose.yml
```diff
@@ -9,7 +9,7 @@ services:
       - db
     volumes:
       - .:/app
-      # - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
+      - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
```
SamGodwin2: Don't forget to comment this back out. :)

repositories/RoutingRepository.js
```diff
@@ -0,0 +1,257 @@
...
+  );
+
+  const { routingDestinationId } = routingRule;
+
+  await checkValidDestination(routingRuleSet.questionPageId, destination);
+  await updateDestination(routingDestinationId, destination);
+  return Routing.updateRoutingRule(
+    id,
+    toDb({
+      operation
```
SamGodwin2: Might want to put a default value here to prevent an empty insert.

repositories/strategies/routingStrategy.js
samiwel: `includes` won't work here since it matches only on value whereas `find` allows you to specify a predicate function or identity.

schema/resolvers.test.js
```diff
@@ -240,7 +240,10 @@ const deleteRoutingConditionMutation = async input =>
     ctx
   );
 
-const updateRoutingConditionValueMutation = async (conditionId, optionId) =>
+const updateRoutingConditionValueMutation = async (
+  conditionId,
+  optionId = null
+) =>
```
samiwel: Nice tidy up

docker-compose.yml
WickyNilliams: @samiwel reminder :)

samiwel:
@WickyNilliams Good shout. I've removed the schema mapping and squashed a few more commits down. If all passes, I reckon we're good to merge. FYI slack connectivity issues for me atm.

WickyNilliams:
@samiwel nice. yeah, getting issues with slack too

PR #78, opened 2018-06-01
Implementing Date Range composite methods.

SamGodwin2:
This PR works to split up a Date range answer into a composite answer type allowing the rest of the application to treat it as two separate answers.

### How to review 
Tests pass and the new Composite type works.

### DIFF
--- a/schema/resolvers.js
+++ b/schema/resolvers.js
@@ -4,12 +4,6 @@ const GraphQLJSON = require("graphql-type-json");
 const formatRichText = require("../utils/formatRichText");
 const getDefaultAnswerProperties = require("../utils/defaultAnswerProperties");
 
-const whereIn = (field, values) => {
-  return function() {
-    this.where("id", "in", values);
-  };
-};
-
 const assertMultipleChoiceAnswer = answer => {
   if (isNil(answer) || !includes(["Checkbox", "Radio"], answer.type)) {
     throw new Error(
@@ -26,9 +20,9 @@ const Resolvers = {
     section: (parent, { id }, ctx) => ctx.repositories.Section.get(id),
     page: (parent, { id }, ctx) => ctx.repositories.Page.get(id),
     questionPage: (_, { id }, ctx) => ctx.repositories.QuestionPage.get(id),
-    answer: (root, { id }, ctx) => ctx.repositories.Answer.get(id),
-    answers: (root, { ids }, ctx) =>
-      ctx.repositories.Answer.findAll(whereIn("id", ids)),
+    answer: (root, { id }, ctx) => ctx.repositories.Answer.getById(id),
+    answers: async (root, { ids }, ctx) =>
+      ctx.repositories.Answer.getAnswers(ids),
     option: (root, { id }, ctx) => ctx.repositories.Option.get(id),
     availableRoutingDestinations: (root, { pageId }, ctx) =>
       ctx.repositories.Routing.getRoutingDestinations(pageId)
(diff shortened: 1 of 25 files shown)


schema/resolvers.js
```diff
@@ -180,17 +189,39 @@ const Resolvers = {
...
 
+  CompositeAnswer: {
+    childAnswers: answer => {
+      const firstAnswer = omit(answer, "secondaryLabel");
+      const secondAnswer = omit(answer, "label");
+      assign(firstAnswer, { id: `${answer.id}from` });
+      assign(secondAnswer, { id: `${answer.id}to` });
+      return [
+        firstAnswer,
+        mapKeys(secondAnswer, (value, key) => {
```
samiwel: Do we need to map keys here? Can't we just set label in the `secondAnswer` in the assign?

schema/resolvers.js
```diff
@@ -115,7 +115,16 @@ const Resolvers = {
 
       return answer;
     },
-    updateAnswer: (_, args, ctx) => ctx.repositories.Answer.update(args.input),
+    updateAnswer: (_, args, ctx) => {
+      if (args.input.id.endsWith("to")) {
+        return ctx.repositories.Answer.update(
+          mapKeys(args.input, (value, key) => {
```
samiwel: Again not sure we need to map all the keys `assign(args.input, { secondaryLabel: args.input.label })`?

schema/resolvers.js
```diff
@@ -115,7 +115,16 @@ const Resolvers = {
 
       return answer;
     },
-    updateAnswer: (_, args, ctx) => ctx.repositories.Answer.update(args.input),
+    updateAnswer: (_, args, ctx) => {
+      if (args.input.id.endsWith("to")) {
```
samiwel: I wonder if this kind of business logic should be pushed down into the Answer repository?

migrations/20180604115603_add_view_for_composite_answers.js
```diff
@@ -0,0 +1,24 @@
+exports.up = function(knex) {
+  return knex.raw(`
+      CREATE OR REPLACE VIEW "SplitAnswersView" AS (
```
samiwel: Is it worth renaming this to be more consistent? `CompositeAnswerView`?

schema/resolvers.js
```diff
@@ -180,17 +189,39 @@ const Resolvers = {
...
+    }
   },
 
   BasicAnswer: {
     page: (answer, args, ctx) =>
       ctx.repositories.QuestionPage.get(answer.questionPageId)
   },
 
+  CompositeAnswer: {
+    childAnswers: answer => {
```
samiwel: might be worth pushing this down into repository?

schema/resolvers.js
adammahmood: Yes, I agree. We should try and keep this module as thin as possible.

adammahmood: Why are we doing a string check against "to"? Doesn't seem future proof.  What about asserting if its a certain type of answer?

scripts/test.sh
```diff
@@ -25,4 +25,4 @@ echo "waiting on postgres to start..."
 
 echo "running tests..."
 
-NODE_ENV=test DB_CONNECTION_URI="postgres://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST/postgres" yarn jest --runInBand $1
+NODE_ENV=test DB_CONNECTION_URI="postgres://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST/postgres" node --inspect node_modules/.bin/jest --runInBand $1
```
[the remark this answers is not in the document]
SamGodwin2: Hmm, not sure how that got into the commit, thought I reverted it, was just a change to get my debugger running.

schema/resolvers.js
SamGodwin2: Ah I think we are forced into it here as well, the front end sends the same object for both child answers, the only difference between them being the id string.

utils/childAnswerParser.js
```diff
@@ -0,0 +1,9 @@
+const { endsWith } = require("lodash/fp");
+
+module.exports = id => {
+  if (endsWith("from", id)) {
+    return "primary";
```
adammahmood: Usually, best to use constant variables rather than strings in these scenarios.  It's ok in this instance as it sounds like we will be re-visiting this area when we build additional composite answers.

PR #88, opened 2018-07-02
Add properties to answers

adammahmood:
This is part of a wider piece of work to introduce answer properties to help the quality of the respondent's input.  

Prior to this work, the concept of answer properties was limited to capturing whether or not an answer is 'mandatory'.  This concept has been refactored to support multiple properties and this PR specifically introduces 'decimal' properties.

[Trello Ticket](https://trello.com/c/f978JRq6/402-properties-decimals-1) 

This PR works in conjunction with the following PR's:

Web App - https://github.com/ONSdigital/eq-author/pull/376
Schema - ONSdigital/eq-author-graphql-schema#46

### How to review 
- Run Tests

### DIFF
--- a/schema/resolvers.js
+++ b/schema/resolvers.js
@@ -1,5 +1,6 @@
 const { GraphQLDate } = require("graphql-iso-date");
-const { includes, isNil } = require("lodash");
+const { merge, includes, isNil } = require("lodash");
+const GraphQLJSON = require("graphql-type-json");
 const formatRichText = require("../utils/formatRichText");
 
 const whereIn = (field, values) => {
@@ -13,6 +14,17 @@ const assertMultipleChoiceAnswer = answer => {
     throw new Error(
       `Answer with id '${answer.id}' must be a Checkbox or Radio.`
     );
+  }
+};
+
+const getDefaultAnswerProperties = type => {
+  switch (type) {
+    case "Currency":
+      return { required: false, decimals: 0 };
+    case "Number":
+      return { required: false, decimals: 0 };
+    default:
+      return { required: false };
   }
 };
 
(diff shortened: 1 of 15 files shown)


adammahmood:
This will require updates to the schema and appropriate version numbers being added to package.json

adammahmood:
[Properties - decimals (1)](https://trello.com/c/f978JRq6/402-properties-decimals-1)

schema/resolvers.js
```diff
@@ -92,7 +102,9 @@ const Resolvers = {
       ctx.repositories.QuestionPage.undelete(args.input.id),
 
     createAnswer: async (root, args, ctx) => {
-      const answer = await ctx.repositories.Answer.insert(args.input);
+      const defaultProperties = getDefaultAnswerProperties(args.input.type);
+      const input = merge({}, args.input, { properties: defaultProperties });
```
samiwel: I think we can get away with not setting default properties on Answer when it's created. When resolving Answer a field resolver for `properties` could do the merge at that point of the default properties and any overridden property values stored on the Answers table.

samiwel: On the other hand, setting properties on creation doesn't lead to unexpected consequences when updating defaults, plus the properties are going to be queried more often than they are created so perhaps makes more sense to leave as is.

schema/resolvers.js
```diff
@@ -16,6 +17,15 @@ const assertMultipleChoiceAnswer = answer => {
   }
 };
 
+const getDefaultAnswerProperties = type => {
+  switch (type) {
+    case "Number":
+      return { required: true, decimals: 0 };
```
samiwel: According to the scenarios, currency should also get a `decimals` property defaulting to zero.

adammahmood: I thought we wanted to do that task separately?

samiwel: @adammahmood Yeah my bad, I probably should have made it explicit on the card. The task is to add a `decimals` property and according to Jimi's scenarios that applies to both `Number` and `Currency` answer types.

AFAIK this is the only place that would need to be changed to introduce it for Currency also? As it's such a small change, I think it makes sense to include it in the same PR.

adammahmood: Yes and probably best to write some cypress tests to validate on the UI.

schema/resolvers.js
```diff
@@ -16,6 +17,15 @@ const assertMultipleChoiceAnswer = answer => {
...
+    default:
+      return { required: true };
```
SamGodwin2: How come we've switched the default case to mandatory/required being true instead of it being false like before?

samiwel:
LGTM, this just needs to be updated to point at schema v0.16.0 once it's been published.

PR #94, opened 2018-07-16
Add move section mutation

adammahmood:

- Implements functionality required to re-order and move sections within a questionnaire

### How to review 
- Run Tests
```
mutation MoveSection($input: MoveSectionInput!) {
  moveSection(input: $input) {
    id
    position
    __typename
  }
}

#Query Variables
{
  "input": {
    "id": "1",
    "questionnaireId": "1",
    "position": 2
  }
}
```

### DIFF
--- a/migrations/20180629130216_add_order_to_sections.js
+++ b/migrations/20180629130216_add_order_to_sections.js
@@ -1,0 +1,14 @@
+exports.up = function(knex) {
+  return knex.schema.table("Sections", table => {
+    table
+      .integer("order")
+      .defaultsTo(0)
+      .notNullable();
+  });
+};
+
+exports.down = function(knex) {
+  return knex.schema.table("Sections", table => {
+    table.dropColumn("order");
+  });
+};
--- a/repositories/SectionRepository.js
+++ b/repositories/SectionRepository.js
@@ -1,42 +1,49 @@
-const { head, invert, map } = require("lodash/fp");
+const { get, head, invert, map, pick } = require("lodash/fp");
 const Section = require("../db/Section");
 const mapFields = require("../utils/mapFields");
 const mapping = { QuestionnaireId: "questionnaireId" };
 const fromDb = mapFields(mapping);
 const toDb = mapFields(invert(mapping));
+const db = require("../db");
+const {
+  getOrUpdateOrderForSectionInsert
+} = require("./strategies/spacedOrderStrategy");
 
 module.exports.findAll = function findAll(
   where = {},
-  orderBy = "created_at",
+  orderBy = "position",
   direction = "asc"
 ) {
-  return Section.findAll()
-    .where({ isDeleted: false })
-    .where(where)
+  return db("SectionsView")
+    .select("*")
+    .where(toDb(where))
     .orderBy(orderBy, direction)
     .then(map(fromDb));
 };
 
 module.exports.getById = function getById(id) {
-  return Section.findById(id)
-    .where({ isDeleted: false })
+  return db("SectionsView")
+    .where("id", parseInt(id, 10))
+    .first()
     .then(fromDb);
 };
 
-module.exports.insert = function insert({
-  title,
-  description,
-  questionnaireId
-}) {
-  return Section.create(
-    toDb({
-      title,
-      description,
-      questionnaireId
-    })
-  )
-    .then(head)
-    .then(fromDb);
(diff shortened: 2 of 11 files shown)


adammahmood:
[re-ordering sections (8)](https://trello.com/c/TAzBTkXp/454-re-ordering-sections-8)

repositories/SectionRepository.js
```diff
@@ -4,39 +4,52 @@ const mapFields = require("../utils/mapFields");
...
+      position
+    )
+      .then(order => Object.assign(args, { order }))
+      .then(({ title, description, questionnaireId, order }) =>
+        toDb({
+          title,
+          description,
+          questionnaireId,
+          order
+        })
```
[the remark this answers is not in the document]
adammahmood: We need to be explicit in what parameters we pass to the field mapper and create section methods.

An alternative solution would be:

```
      .then(order => Object.assign(args, { order }))
      .then(pick(['title', 'description', 'questionnaireId', 'order']))
      .then(toDb)
      .then(section => Section.create(section, trx))
      .then(head)
      .then(fromDb);
```
Which solution do you think is clearer?

repositories/strategies/spacedOrderStrategy.js
```diff
@@ -46,13 +58,48 @@ const getOrUpdateOrderForInsert = async (trx, sectionId, id, position) => {
...
+  }
+  if (position === sections.length) {
+    return maxOrder;
+  }
+
+  let left = getOr(0, "order", sections[position - 1]);
+  let right = getOr(maxOrder, "order", sections[position]);
+
+  if (valuesHaveConverged(left, right)) {
+    await makeSpaceForSectionInsert(trx, questionnaireId, left);
```
samiwel: There's some duplication of logic here that could be eliminated.

```js
const makeSpaceForInsert = (trx, tableName, parentId, order) => {
  const key = tableName === "Sections" ? "QuestionnaireId" : "SectionId";
  return trx(tableName)
    .where({ [key]: parentId })
    .andWhere("order", ">", order)
    .increment("order", SPACING);
};

const getOrUpdateOrderForInsert = async (
  trx,
  collection,
  type,
  parentId,
  position
) => {
  const maxOrder = getMaxOrder(collection) + SPACING;

  position = isNil(position)
    ? collection.length
    : clamp(0, collection.length, position);

  if (isEmpty(collection)) {
    return SPACING;
  }
  if (position === collection.length) {
    return maxOrder;
  }

  let left = getOr(0, "order", collection[position - 1]);
  let right = getOr(maxOrder, "order", collection[position]);

  if (valuesHaveConverged(left, right)) {
    await makeSpaceForInsert(trx, type, parentId, left);
    right += SPACING;
  }

  return calculateMidPoint(left, right);
};

const getOrUpdateOrderForPageInsert = async (trx, sectionId, id, position) => {
  const pages = reject({ id }, await getPagesBySection(trx, sectionId));

  return getOrUpdateOrderForInsert(trx, pages, "Pages", sectionId, position);
};

const getOrUpdateOrderForSectionInsert = async (
  trx,
  questionnaireId,
  id,
  position
) => {
  const sections = reject(
    { id },
    await getSectionsByQuestionnaire(trx, questionnaireId)
  );

  return getOrUpdateOrderForInsert(
    trx,
    sections,
    "Sections",
    questionnaireId,
    position
  );
};
```

samiwel: You could possibly get rid of `getOrUpdateOrderForPageInsert` and `getOrUpdateOrderForSectionInsert` altogether.

samiwel:
In general looks good to me failing on code coverage check though, so might want to look at getting that coverage back up?

migrations/20180629130216_add_order_to_sections.js
```diff
@@ -0,0 +1,14 @@
+exports.up = function(knex) {
+  return knex.schema.table("Sections", table => {
+    table
+      .integer("order")
+      .defaultsTo(0)
+      .notNullable();
+  });
```
SamGodwin2: Think we need to retroactively add section order to previously created sections here as otherwise all sections that already exist will suddenly have an order of 0.

adammahmood: New migration added 'seed_section_order_columns'

samiwel:
LGTM, just did some testing in graphiql seems to be doing the right thing 👍

PR #97, opened 2018-08-01
Add query resolver to get total number of sections in a questionnaire

adammahmood:
- Resolves new type `QuestionnaireInfo` which returns total number of sections in a questionnaire
- This is needed to be able to disable/enable the 'Move Section' button on the UI to based on the number of sections in a questionnaire.
- https://trello.com/c/18p5lgJP/533-disable-move-section-button-if-there-is-only-one-section

### How to review
- Run Tests
- Use eQ Author GraphQL API explorer to query QuestionnaireInfo eg.
```
query GetTotalSectionCount($id: ID!) {
  questionnaire(id: $id) {
    questionnaireInfo {
      totalSectionCount
    }
  }
}

#Query Variables
{
  "id": 1
}
```

### DIFF
--- a/repositories/SectionRepository.test.js
+++ b/repositories/SectionRepository.test.js
@@ -114,6 +114,17 @@ describe("SectionRepository", () => {
     expect(position).toEqual("0");
   });
 
+  it("can get section count ", async () => {
+    const {
+      questionnaire: { id: questionnaireId }
+    } = await setup();
+
+    await SectionRepository.insert(buildSection({ questionnaireId }));
+
+    const count = await SectionRepository.getSectionCount(questionnaireId);
+    expect(count).toEqual("1");
+  });
+
   describe("re-ordering", () => {
     const createSections = (questionnaireId, numberOfPages) => {
       const sections = times(numberOfPages, i =>
--- a/schema/resolvers.js
+++ b/schema/resolvers.js
@@ -175,7 +175,13 @@ const Resolvers = {
   Questionnaire: {
     sections: (questionnaire, args, ctx) =>
       ctx.repositories.Section.findAll({ QuestionnaireId: questionnaire.id }),
-    createdBy: questionnaire => ({ name: questionnaire.createdBy })
+    createdBy: questionnaire => ({ name: questionnaire.createdBy }),
+    questionnaireInfo: ({ id }) => id
+  },
+
+  QuestionnaireInfo: {
+    totalSectionCount: (questionnaireId, args, ctx) =>
+      ctx.repositories.Section.getSectionCount(questionnaireId)
   },
 
   Section: {
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "colors": "^1.1.2",
     "cors": "^2.8.3",
     "dotenv": "^6.0.0",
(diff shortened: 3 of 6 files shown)


adammahmood:
[Disable Move Section button if there is only one section](https://trello.com/c/18p5lgJP/533-disable-move-section-button-if-there-is-only-one-section)

repositories/SectionRepository.test.js
```diff
@@ -114,6 +114,19 @@ describe("SectionRepository", () => {
     expect(position).toEqual("0");
   });
 
+  it("can get section count ", async () => {
+    const {
+      questionnaire: { id: questionnaireId }
+    } = await setup();
+
+    await SectionRepository.insert(
+      buildSection({ questionnaireId: questionnaireId })
```
samiwel: This could just be

```js
buildSection({ questionnaireId })
```

repositories/SectionRepository.test.js
```diff
@@ -114,6 +114,19 @@ describe("SectionRepository", () => {
...
+    );
+
+    const count = await SectionRepository.getSectionCount(questionnaireId);
+    expect(count).toEqual("1");
```
samiwel: Why is 1 a string as opposed to a number?

schema/resolvers.js
```diff
@@ -175,7 +175,13 @@ const Resolvers = {
   Questionnaire: {
     sections: (questionnaire, args, ctx) =>
       ctx.repositories.Section.findAll({ QuestionnaireId: questionnaire.id }),
-    createdBy: questionnaire => ({ name: questionnaire.createdBy })
+    createdBy: questionnaire => ({ name: questionnaire.createdBy }),
+    questionnaireInfo: ({ id }) => id
```
samiwel: Is `questionnaireInfo` field required here? If it's omitted would the `QuestionnaireInfo` resolver still work like this?

```js
QuestionnaireInfo: {
    totalSectionCount: ({id: questionnaireId}, args, ctx) =>
      ctx.repositories.Section.getSectionCount(questionnaireId)
}
```

repositories/SectionRepository.test.js
adammahmood: Good point. Will update.

repositories/SectionRepository.test.js
adammahmood: From Knex docs

```
count — .count(column|columns|raw)

Performs a count on the specified column or array of columns (note that some drivers do not support multiple columns). Also accepts raw expressions. Note that in *Postgres*, count returns a bigint type which will be a String and not a Number (more info).
```

adammahmood: Postgres returns String as count value

schema/resolvers.js
adammahmood: If omitted it doesn't resolve to `QuestionnaireInfo` and response is:

```
{
  "data": {
    "section": {
      "id": "4",
      "title": "<p>S1</p>",
      "description": "",
      "position": 0,
      "__typename": "Section",
      "questionnaire": {
        "questionnaireInfo": null
      }
    }
  }
}
```

PR #102, opened 2018-08-08
Validation min and max value API changes

SamGodwin2:
This PR implements the changes required in order to mutate and retrieve data regarding minimum and maximum value mutations. It includes new resolvers, new strategies and methods in the repositories and finally also includes db mutations required to add the minValue and maxValue data to our db schema.  

### How to review 
Tests all pass, and using the graphiql interface all minValue and maxValue data should be retrievable.

### DIFF
--- a/repositories/AnswerRepository.js
+++ b/repositories/AnswerRepository.js
@@ -6,7 +6,10 @@ const {
   invert,
   map,
   omit,
-  assign
+  assign,
+  values,
+  includes,
+  flatten
 } = require("lodash/fp");
 const { get, merge } = require("lodash");
 const db = require("../db");
@@ -14,6 +17,11 @@ const Answer = require("../db/Answer");
 const mapFields = require("../utils/mapFields");
 const mapping = { QuestionPageId: "questionPageId" };
 const childAnswerParser = require("../utils/childAnswerParser");
+const getDefaultAnswerProperties = require("../utils/defaultAnswerProperties");
+const {
+  createDefaultValidationsForAnswer
+} = require("../repositories/strategies/validationStrategy");
+const { answerTypeMap } = require("../utils/defaultAnswerValidations");
 
 const handleDeprecatedMandatoryFieldFromDb = answer =>
   isObject(answer)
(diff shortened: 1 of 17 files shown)


samiwel:
[Validation - Min value API changes](https://trello.com/c/SlGVou7h/500-validation-min-value-api-changes)

utils/defaultAnswerValidations.js
```diff
@@ -0,0 +1,22 @@
+const answerTypeMap = {
+  number: ["Currency", "Number"]
+};
+
+const validationFields = {
```
samiwel: Can we rename to `validationRuleMap` or something. fields is a little confusing given GraphQL has fields

utils/defaultAnswerValidations.js
```diff
@@ -0,0 +1,22 @@
...
+  number: ["minValue", "maxValue"]
+};
+
+const validationConfigs = {
```
samiwel: `defaultValdationRuleConfig`?

repositories/AnswerRepository.js
```diff
@@ -55,30 +63,72 @@ const getById = id =>
...
-      questionPageId
-    })
-  )
+const createAnswer = async (args, ctx) => {
+  return db.transaction(async trx => {
+    const defaultProperties = getDefaultAnswerProperties(args.input.type);
+    const input = merge({}, args.input, { properties: defaultProperties });
+    const answer = await ctx.repositories.Answer.insert(input, trx);
+
+    if (includes(answer.type, flatten(values(answerTypeMap)))) {
```
samiwel: Going forward I don't think the flatten values approach will work.

Right now you have:

```js
{
  number: ["Number", "Currency"]
}
```

I think this will need to be flipped to something like this:

```js
{
 Number: "number",
 Currency: "number"
}
```

Then you look up which set of validation rules apply with `answerTypeMap[answer.type]`.

Happy to leave this as is for now, because it works in the case where we have a single validation type. But this will need to change when we add date and duration validations.

repositories/ValidationRepository.js
```diff
@@ -0,0 +1,29 @@
+const Validation = require("../db/Validation");
+const { head, flow, keys, remove, first } = require("lodash/fp");
+
+const toggleValidationRule = ({ id, enabled }) => {
+  return Validation.update(id, { enabled }).then(head);
+};
+
+const findByTypeId = ({ id }, validationType) => {
```
samiwel: Sorry to be pedantic but this is really `findByAnswerIdAndValidationType`

schema/resolvers.js
```diff
@@ -348,6 +327,57 @@ const Resolvers = {
       ctx.repositories.Answer.getById(answerId)
   },
 
+  ValidationType: {
+    __resolveType: answer => {
+      const validationEntity = getValidationEntity(answer.type);
+
+      switch (validationEntity) {
+        case "number":
+          return "NumberValidation";
```
samiwel: Going back to my first comment about lookup of validation type from the answer type we could eliminate the need to switch here altogether if we had something like:

```json
{
  "Number": "NumberValidation",
  "Currency": "NumberValidation"
}
```

Here we could just do this (and getValidationEntity would become redundant):

```js
return answerTypeMap[answer.type];
```

samiwel: Again probably leave it for now, but worth noting for future refactoring.

schema/resolversTests/validation.test.js
```diff
@@ -0,0 +1,196 @@
...
+        inclusive: true
+      }
+    });
+    expect(result).toMatchObject({
+      id: currencyValidation.maxValue.id,
+      custom: 10,
+      inclusive: true
+    });
+  });
+});
```
samiwel: Good that we've separated this out I think 👍

samiwel:
Also looks like build is failing

repositories/AnswerRepository.js
SamGodwin2: This values flatten only checks that the answer type is a valid type for validations so it should create an array of all accepted answer types. Then if the current answer type is included in those it goes to the create default validations where it uses a couple different lodash functions to work out what validations to put in.

I could be wrong but I *think* this approach should still work with new validations, but it might be too complex just for the sake of keeping the default object a tiny bit smaller.

PR #106, opened 2018-08-20
Allow non-number answers to return null from validation resolver.

adammahmood:
Returns null from the validation node in basic answers for all non-number answer types.

### How to review
Run tests


samiwel:
As discussed, introducing new type is not required just return null for validation if the answer is not a number type.

PR #107, opened 2018-08-23
Added new migration

SamGodwin2:
The changes made to the api during the addition of min value validation were not backwards compatible as no data was being created for the existing answers.
This PR aims to add a mutation which can add empty data to all existing answers that now need those validation rows.  

### How to review 
Test on database, add number answer, delete the validations, then run this migration.

Test using a large amount of data, quickest way I found to create the data was to use this mutation on the graphql client.

mutation{
  createAnswer(input: {
    type: Currency
    questionPageId: 1
  }){
   id 
  }
}

Then execute the following code in the browser:

` var b = document.querySelector("[class='execute-button']")`
`for (var i =0 ; i<2000; i++){
    b.click()
}`

Finally delete everything in the validations database and run the mutation.

### DIFF
--- a/migrations/20180822122408_add_min_max_validation_to_existing_number_currency_answers.js
+++ b/migrations/20180822122408_add_min_max_validation_to_existing_number_currency_answers.js
@@ -1,0 +1,29 @@
+exports.up = async function(knex) {
+  const ids = await knex
+    .select("Answers.id")
+    .from("Answers")
+    .where(function() {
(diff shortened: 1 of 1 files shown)


SamGodwin2:
[Migration to allow the min validation changes to be backwards compatible.](https://trello.com/c/ThHXLDFR/578-migration-to-allow-the-min-validation-changes-to-be-backwards-compatible)

PR #109, opened 2018-08-28
Mutually exclusive answer options

SamGodwin2:
This Pr adds the required resolvers, methods, migrations and tests to ensure that mutually exclusive answers can be created in the database. 

### How to review 
Tests pass and are conclusive, exclusive answers can be created on the back end.

### Dependency
https://github.com/ONSdigital/eq-author-graphql-schema/pull/56

### DIFF
--- a/repositories/OptionRepository.js
+++ b/repositories/OptionRepository.js
@@ -7,6 +7,19 @@ const mapping = { AnswerId: "answerId" };
 const fromDb = mapFields(mapping);
 const toDb = mapFields(invert(mapping));
 
+const findExclusiveOptionByAnswerId = answerId =>
+  Option.findAll()
+    .where(
+      toDb({
+        isDeleted: false,
+        otherAnswerId: null,
+        mutuallyExclusive: true,
+        answerId
+      })
+    )
+    .then(head)
+    .then(fromDb);
+
 const checkForExistingExclusive = async answerId => {
   const existingExclusive = await findExclusiveOptionByAnswerId(answerId);
   if (!isNil(existingExclusive)) {
@@ -20,19 +33,6 @@ const findAll = (where = {}, orderBy = "created_at", direction = "asc") =>
     .where(where)
     .orderBy(orderBy, direction)
     .then(map(fromDb));
-
-const findExclusiveOptionByAnswerId = answerId =>
-  Option.findAll()
-    .where(
-      toDb({
-        isDeleted: false,
-        otherAnswerId: null,
-        mutuallyExclusive: true,
-        answerId
-      })
-    )
-    .then(head)
-    .then(fromDb);
 
 const getById = id =>
   Option.findById(id)
--- a/repositories/OptionRepository.js
+++ b/repositories/OptionRepository.js
@@ -1,4 +1,4 @@
-const { head, invert, map } = require("lodash/fp");
+const { head, invert, map, isEmpty } = require("lodash/fp");
 const Option = require("../db/Option");
 const mapFields = require("../utils/mapFields");
 const db = require("../db");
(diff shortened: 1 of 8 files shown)


SamGodwin2:
[Mutually exclusive answer options - Backend](https://trello.com/c/23VKl7hj/567-mutually-exclusive-answer-options-backend)

docker-compose.yml
```diff
@@ -9,7 +9,7 @@ services:
       - db
     volumes:
       - .:/app
-      #- ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
```
adammahmood: Best to use `git add -p` to select what to stage and commit to prevent this.

repositories/OptionRepository.js
```diff
@@ -7,6 +7,19 @@ const mapping = { AnswerId: "answerId" };
...
+const checkForExistingExclusive = async answerId => {
+  const existingExclusive = await Option.findAll().where(
+    toDb({
+      answerId,
+      mutuallyExclusive: true,
+      isDeleted: false
+    })
+  );
+  if (!isEmpty(existingExclusive)) {
+    throw new Error("There is already a exclusive checkbox on this answer.");
```
adammahmood: Typo.  This should read: `There is already an exclusive checkbox on this answer`

repositories/OptionRepository.js
```diff
@@ -19,24 +32,41 @@ module.exports.findAll = function findAll(
     .then(map(fromDb));
 };
 
+module.exports.findOne = function findOne(
```
adammahmood: This should be named `getById` to follow the convention of other repositories.

repositories/OptionRepository.js
```diff
@@ -7,6 +7,19 @@ const mapping = { AnswerId: "answerId" };
 const fromDb = mapFields(mapping);
 const toDb = mapFields(invert(mapping));
 
+const checkForExistingExclusive = async answerId => {
```
adammahmood: We should unit test this.

repositories/OptionRepository.js
SamGodwin2: This isn't quite the same functionality of getById, as I needed a function where I could pass it a number of different conditions and have it return only the first result. I suppose the best option here would be to specialise it further and name it getExclusiveOption. Plus we already have a getById in here.

repositories/OptionRepository.js
SamGodwin2: Do we need to? The pattern we've used in other places suggests that testing this in the resolvers integration test should be fine?

repositories/OptionRepository.js
adammahmood: Sorry.  Mis-read the function. This is identical to `findAll()` apart from the `head()` right at the end.  I would either utilise the `findAll()` method to prevent duplication or leave the resposibility of `head()` to the caller.

SamGodwin2: Not sure if you saw the change but I ended up specialising the function so it only relates to Exclusives now, makes it a bit easier to call as only the answerId needs passing to it plus it makes the function itself simpler.

repositories/OptionRepository.js
```diff
@@ -32,16 +32,19 @@ module.exports.findAll = function findAll(
     .then(map(fromDb));
 };
 
-module.exports.findOne = function findOne(
-  where = {},
-  orderBy = "created_at",
-  direction = "asc"
+module.exports.findExclusiveOptionByAnswerId = function findExclusiveOptionByAnswerId(
```
adammahmood: I still think this is unnecessary as we can just utilise `findAll()`

SamGodwin2: I disagree here, I'd rather have a specialised function for something like this, as I'd like to make it very obvious what it's doing when we're calling it from the resolvers file, plus it just allows us to keep the resolver functions dumber.

adammahmood: I understand your point of view.  However, I would prefer not to pollute the repositories with lots of wrapper functions. 

Nevertheless, I'm happy to resolve this conversation and we can discuss in the future if it becomes problematic.

PR #111, opened 2018-08-30
561 create mutation for duplicating a page

K9Morphed:
This PR adds the functionality in order to duplicate a page. 

>It also includes code for duplicating a section and questionnaire; these functions will not work just yet because some implementation for duplicating a page breaks these. This should be fixed in separate PRs because the additional functionality is not in scope for this ticket.

### How to review 
All tests should pass, and the code should work when calling the mutations included in [this schema PR](https://github.com/ONSdigital/eq-author-graphql-schema/pull/57)

### DIFF
--- a/repositories/OptionRepository.js
+++ b/repositories/OptionRepository.js
@@ -27,7 +27,7 @@ const checkForExistingExclusive = async answerId => {
   }
 };
 
-const findAll = (where = {}, orderBy = "created_at", direction = "asc") =>
+const findAll = (where = {}, orderBy = "id", direction = "asc") =>
   Option.findAll()
     .where({ isDeleted: false, otherAnswerId: null })
     .where(where)
--- a/repositories/PageRepository.js
+++ b/repositories/PageRepository.js
@@ -1,7 +1,9 @@
 const { map, head, invert, get } = require("lodash/fp");
+const updateTitle = require("../utils/updateTitle");
 const Page = require("../db/Page");
 const QuestionPageRepository = require("./QuestionPageRepository");
 const mapFields = require("../utils/mapFields");
+const { duplicatePageStrategy } = require("./strategies/duplicateStrategy");
 const db = require("../db");
 const {
   getOrUpdateOrderForPageInsert
@@ -95,6 +97,20 @@ function getRoutingDestinations(pageId) {
   return db.transaction(trx => getAvailableRoutingDestinations(trx, pageId));
 }
 
+function duplicatePage(id, position) {
+  return db.transaction(async trx => {
+    const page = await trx
+      .select("*")
+      .from("Pages")
+      .where({ id })
+      .then(head);
+
+    return duplicatePageStrategy(trx, page, position, {
+      title: updateTitle(page.title)
+    }).then(fromDb);
+  });
+}
+
 Object.assign(module.exports, {
   findAll,
   getById,
@@ -104,5 +120,6 @@ Object.assign(module.exports, {
   undelete,
   move,
   getPosition,
-  getRoutingDestinations
+  getRoutingDestinations,
+  duplicatePage
 });
--- a/repositories/PageRepository.test.js
+++ b/repositories/PageRepository.test.js
@@ -2,7 +2,7 @@ const db = require("../db");
 const QuestionnaireRepository = require("../repositories/QuestionnaireRepository");
 const SectionRepository = require("../repositories/SectionRepository");
 const PageRepository = require("../repositories/PageRepository");
-const { last, head, map, times } = require("lodash");
+const { last, head, map, times, omit, parseInt } = require("lodash");
 
 const reverse = array => array.slice().reverse();
 
@@ -52,7 +52,7 @@ const eachP = (items, iter) =>
 
 describe("PagesRepository", () => {
   beforeAll(() => db.migrate.latest());
-  afterAll(() => db.migrate.rollback().then(() => db.destroy()));
+  afterAll(() => db.destroy());
   afterEach(() => db("Questionnaires").delete());
 
   it("throws for unknown page types", () => {
@@ -194,7 +194,10 @@ describe("PagesRepository", () => {
     });
 
     it("can move pages between sections", async () => {
-      const { section, questionnaire: { id: questionnaireId } } = await setup();
+      const {
+        section,
+        questionnaire: { id: questionnaireId }
+      } = await setup();
 
       const section2 = await SectionRepository.insert(
         buildSection({ title: "Section 2", questionnaireId })
@@ -215,7 +218,10 @@ describe("PagesRepository", () => {
     });
 
     it("correctly re-orders pages as they're moved between sections", async () => {
-      const { section, questionnaire: { id: questionnaireId } } = await setup();
+      const {
+        section,
+        questionnaire: { id: questionnaireId }
+      } = await setup();
 
       const section2 = await SectionRepository.insert(
         buildSection({ title: "Section 2", questionnaireId })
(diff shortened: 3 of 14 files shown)


repositories/PageRepository.test.js
```diff
@@ -440,4 +446,73 @@ describe("PagesRepository", () => {
...
+      const pageOne = await PageRepository.insert(
+        buildPage({ sectionId: section.id })
+      );
+
+      const pageTwo = await PageRepository.copyPage(pageOne.id, 1);
+
+      const positionOfPageOne = await PageRepository.getPosition(pageOne);
+      const positionOfPageTwo = await PageRepository.getPosition(pageTwo);
+
+      expect(parseInt(positionOfPageOne) + 1).toEqual(
```
adammahmood: It's a bit annoying that we have to use `parseInt` on the `position` values all the time.  I have created a Tech Debt ticket to investigate this: https://trello.com/c/r1NbxAwr/312-review-bigint-usage-in-api

repositories/strategies/duplicateStrategy.js
```diff
@@ -0,0 +1,230 @@
...
+  if (!isNil(parentRelation)) {
+    duplicatedRecord[get(parentRelation, "columnName")] = get(
+      parentRelation,
+      "id"
+    );
+  }
+
+  const newRecord = { ...duplicatedRecord, ...other };
+
+  return Promise.resolve(
```
adammahmood: The `Promise.resolve` is unnecessary.  This pattern also appears frequently in this file.

repositories/strategies/duplicateStrategy.js
```diff
@@ -0,0 +1,230 @@
+const { omit, head, get, isNil } = require("lodash");
```
adammahmood: This file should be unit tested.

repositories/strategies/duplicateStrategy.js
```diff
@@ -0,0 +1,230 @@
...
+
+  return returnedData;
+};
+
+const selectData = async (trx, tableName, columns, where, callback) => {
+  return trx
+    .select(columns)
+    .from(tableName)
+    .where(where)
+    .then(callback);
```
adammahmood: Unnecessary callback.  The promise is returned so the caller can add their own `.then()`

schema/resolvers.js
```diff
@@ -47,6 +47,8 @@ const Resolvers = {
       ctx.repositories.Questionnaire.remove(args.input.id),
     undeleteQuestionnaire: (_, args, ctx) =>
       ctx.repositories.Questionnaire.undelete(args.input.id),
+    duplicateQuestionnaire: (_, args, ctx) =>
```
adammahmood: We should add a resolver test for the new functions in this file.

repositories/PageRepository.test.js
K9Morphed: @adammahmood Alright. I'll resolve this conversation for now and then we can revisit this once the TechDebt card has been completed, if you agree? I don't feel that it should be a blocker for this task.

repositories/PageRepository.js
```diff
@@ -1,7 +1,9 @@
 const { map, head, invert, get } = require("lodash/fp");
+const replaceTitle = require("../utils/updateTitle");
```
samiwel: Name of function and file should ideally be the same.

repositories/PageRepository.js
```diff
@@ -95,6 +97,20 @@ function getRoutingDestinations(pageId) {
   return db.transaction(trx => getAvailableRoutingDestinations(trx, pageId));
 }
 
+function copyPage(id, position) {
```
samiwel: Rename to `duplicatePage` so that our nomenclature is consistent?

repositories/PageRepository.js
```diff
@@ -95,6 +97,20 @@ function getRoutingDestinations(pageId) {
...
+  return db.transaction(async trx => {
+    const page = await trx
+      .select("*")
+      .from("Pages")
+      .where({ id: id })
```
samiwel:
```js
.where({ id })
```

repositories/QuestionnaireRepository.js
```diff
@@ -3,6 +3,24 @@ const Questionnaire = require("../db/Questionnaire");
 const mapFields = require("../utils/mapFields");
 const mapping = { created_at: "createdAt" }; // eslint-disable-line camelcase
 const fromDb = mapFields(mapping);
+const db = require("../db");
+const {
+  duplicateQuestionnaireStrategy
+} = require("./strategies/duplicateStrategy");
+
+module.exports.copyQuestionnaire = function(id) {
```
samiwel: Same as above inconsistent naming

utils/updateTitle.js
```diff
@@ -0,0 +1,6 @@
+const { replace, isNil } = require("lodash");
+
+module.exports = (title, prefix = "Copy of") =>
+  replace(title, /(<\w*>)?(.*)/, (_, a, b) => {
+    return isNil(a) ? `${prefix} ${b}` : `${a}${prefix} ${b}`;
+  });
```
samiwel: This would benefit from some unit tests. Be good to test passing a title surrounded by tags, containing tags and without tags so behaviour is documented by example.

repositories/PageRepository.test.js
adammahmood: I agree.

repositories/PageRepository.js
adammahmood: Great use of the word "nomenclature".

samiwel:
As discussed duplicated options appear out of order. Other than that looks good and I've been testing with my front end changes seems pretty solid 👍

tgandrews:
@samiwel this should be resolved now.

repositories/OptionRepository.js
```diff
@@ -27,7 +27,7 @@ const checkForExistingExclusive = async answerId => {
   }
 };
 
-const findAll = (where = {}, orderBy = "created_at", direction = "asc") =>
+const findAll = (where = {}, orderBy = "id", direction = "asc") =>
```
adammahmood: Is ordering by `id` safe? Why doesn't order by `created_at` at work?  If the `created_at` becomes the same value when duplicating. Then maybe we should order by `created_at` and then fall back onto `id`?

repositories/strategies/duplicateStrategy.js
```diff
@@ -0,0 +1,214 @@
...
+    const order = await getOrUpdateOrderForPageInsert(
+      trx,
+      returnedData.SectionId,
+      returnedData.id,
+      position
+    );
+
+    await trx
+      .table(tableName)
+      .update({ order: order })
```
adammahmood: This could be `.update({order})`

adammahmood: I've seen (and done this myself) a few times.  Maybe we can set up a linting rule to enforce this?

scripts/test.sh
```diff
@@ -25,4 +25,4 @@ echo "waiting on postgres to start..."
 
 echo "running tests..."
 
-NODE_ENV=test DB_CONNECTION_URI="postgres://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST/postgres" yarn jest --runInBand $1
+NODE_ENV=test DB_CONNECTION_URI="postgres://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_HOST/postgres" yarn jest --runInBand "$@"
```
adammahmood: This should _really_ be a separate commit.

repositories/OptionRepository.js
tgandrews: When we insert in during a duplication the `created_at` date is all the same so the order could not be determined. This resolved that issue. I don't see a point in having a fall back if `id` works just as well as `created_at`.

repositories/strategies/duplicateStrategy.js
tgandrews: Yeah we can add it - I have definitely seen it before. I think airbnb eslint does it by default. But its a tech debt.

scripts/test.sh
tgandrews: Yes apologies. This was done debugging and I forgot to split it out. It just passes all args passed to the script down to `jest` not just the first one.

repositories/OptionRepository.js
samiwel: Ordering by `id` should be safe enough until we have a requirement to re-order options.

scripts/test.sh
adammahmood: Can we pull this out into a seperate commit?

adammahmood:
LGTM.  Squash commits and ensure the enhancement to `scripts/test.sh` is a separate commit to the feature commit.

PR #116, opened 2018-09-05
Fixed bug preventing move pages/sections from working

SamGodwin2:
There is currently a bug on the move page api where trying to move a page forward creates unpredictable results this PR aims to bring the move page feature back into a working state.

### How to review 
New test passes and the move page feature works.

### DIFF
--- a/repositories/PageRepository.test.js
+++ b/repositories/PageRepository.test.js
@@ -139,7 +139,7 @@ describe("PagesRepository", () => {
       });
     });
 
-    it("can move pages within same section", async () => {
+    it("can move pages backwards within same section", async () => {
       const { section } = await setup();
       const pages = await createPages(section.id, 5);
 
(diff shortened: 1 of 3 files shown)


repositories/PageRepository.test.js
```diff
@@ -159,6 +159,27 @@ describe("PagesRepository", () => {
       expect(map(updatePages, "id")).toEqual(map(reverse(pages), "id"));
     });
 
+    it("can move pages forwards within same section", async () => {
+      const { section } = await setup();
+      const pages = await createPages(section.id, 5);
+
+      const middlePage = nth(pages, 3);
```
tgandrews: any reason we aren't just using `pages[3]`?

repositories/strategies/spacedOrderStrategy.js
```diff
@@ -60,8 +60,10 @@ const getOrUpdateOrderForInsert = async (
     return maxOrder;
   }
 
+  collection.splice(position, 0, { id: "dummyElement" });
```
tgandrews: Maybe this work could be done as an abstract util for array moving so it can be tested as a pure function

SamGodwin2:
[Unable to correctly move pages forwards through the section.](https://trello.com/c/bE8RFau9/594-unable-to-correctly-move-pages-forwards-through-the-section)

repositories/strategies/spacedOrderStrategy.js
SamGodwin2: I do agree here but after looking at it I'm not sure it would work as a pure function as it requires the use of database functions when the values on either side converge. We could pull that function out and then make the rest a pure function but I feel that might make a already complicated function even more complex?

tgandrews: Yeah I was hoping to make just the array manipulation pure. So it would prep the data ready but if that’s not possible it can wait as it isn’t too bad.

repositories/PageRepository.test.js
```diff
@@ -159,6 +159,27 @@ describe("PagesRepository", () => {
       expect(map(updatePages, "id")).toEqual(map(reverse(pages), "id"));
     });
 
+    it("can move pages forwards within same section", async () => {
+      const { section } = await setup();
+      const pages = await createPages(section.id, 5);
+
+      const middlePage = pages[3];
+
+      // reverse the list
```
samiwel: Spurious comment?

repositories/PageRepository.test.js
```diff
@@ -159,6 +159,27 @@ describe("PagesRepository", () => {
...
+        position: "5"
+      });
+
+      const updatePages = await PageRepository.findAll({
+        sectionId: section.id
+      });
+
+      expect(last(updatePages).id).toEqual(middlePage.id);
+    });
+
```
samiwel: Worth writing another test moving forward but not to the end?

PR #117, opened 2018-09-05
Rename all foreign key columns in Postgres to use camel case

SamGodwin2:
There is currently a pain point in the api where fk columns are pascal cased however the code written is in a camel case format. This means that whenever we want to interact with the database we need to run the passed parameters through a toDB/fromDB function to convert the format. 
This PR aims to rename all fk columns to camel case to allow us to remove this confusion.

### How to review 
All api tests pass and the api/front-end interact as normal.


test.txt
```diff
@@ -0,0 +1,4253 @@
+[999D[K[999D[K[0m[7m[1m[31m FAIL [39m[22m[27m[0m [2mschema/[22m[1mresolvers.test.js[22m
```
tgandrews: I presume this has been unintentionally committed?

SamGodwin2: Whoops!

SamGodwin2:
[Investigate renaming of columns in Postgres](https://trello.com/c/D0IDEKxF/587-investigate-renaming-of-columns-in-postgres)

migrations/20180903142820_renaming_fk_columns_in_db.js
```diff
@@ -0,0 +1,154 @@
...
+  promiseArray.push(
+    knex.schema.table("Metadata", t => {
+      t.renameColumn("QuestionnaireId", "questionnaireId");
+    })
+  );
+
+  return Promise.all(promiseArray);
+};
+
+exports.down = function(knex) {
```
tgandrews: Do we need this?

migrations/20180907154321_recreate_views_with_new_naming_conventions.js
```diff
@@ -0,0 +1,66 @@
+exports.up = async function(knex, Promise) {
+  await knex.raw(`DROP VIEW "PagesView"`);
```
tgandrews: Do we need to drop the view given we are using `CREATE OR REPLACE VIEW`?

SamGodwin2: That's what I thought but for some weird reason, if I try to use `create or replace` without dropping the views first I get an error in the migration preventing it from happening. It just complains saying I can't rename the columns...

migrations/20180903142820_renaming_fk_columns_in_db.js
SamGodwin2: Probably not, pretty much just added it for the sake of consistency, worth having that tech session soon so we can decide whether to be rid of them for good.

migrations/20180907154321_recreate_views_with_new_naming_conventions.js
tgandrews: That's daft 😞

PR #120, opened 2018-09-12
Fix reference to historic foreign key in database

adammahmood:
Fixes reference to a historic foreign key in the database

### How to review 
- Run Tests

### DIFF
--- a/schema/resolvers.js
+++ b/schema/resolvers.js
@@ -171,7 +171,7 @@ const Resolvers = {
     createdBy: questionnaire => ({ name: questionnaire.createdBy }),
     questionnaireInfo: ({ id }) => id,
     metadata: (questionnaire, args, ctx) =>
-      ctx.repositories.Metadata.findAll({ QuestionnaireId: questionnaire.id })
+      ctx.repositories.Metadata.findAll({ questionnaireId: questionnaire.id })
   },
 
   QuestionnaireInfo: {
--- a/tests/schema/queries/metadata.test.js
+++ b/tests/schema/queries/metadata.test.js
@@ -48,7 +48,7 @@ describe("Metadata query", () => {
     );
     expect(result.errors).toBeUndefined();
     expect(repositories.Metadata.findAll).toHaveBeenCalledWith({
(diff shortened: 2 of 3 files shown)


schema/resolvers.js
```diff
@@ -171,7 +171,7 @@ const Resolvers = {
     createdBy: questionnaire => ({ name: questionnaire.createdBy }),
     questionnaireInfo: ({ id }) => id,
     metadata: (questionnaire, args, ctx) =>
-      ctx.repositories.Metadata.findAll({ QuestionnaireId: questionnaire.id })
+      ctx.repositories.Metadata.findAll({ questionnaireId: questionnaire.id })
```
tgandrews: Does this requires a test given it was missed in the previous PR?

tests/schema/queries/metadata.test.js
```diff
@@ -48,7 +48,7 @@ describe("Metadata query", () => {
     );
     expect(result.errors).toBeUndefined();
     expect(repositories.Metadata.findAll).toHaveBeenCalledWith({
```
tgandrews: How was it passing?

tgandrews: Should we add an integration test for this? Or improve the api so it isn't passing db fields in?

adammahmood: I think it would be better to have an integration test.  I'll get one written.

tgandrews: I feel it just is showing that the repository api is leaky - you need to know the internals.

adammahmood: The best implementation would be to have an 'entity' model which would represent your data model.  This would then hide any implementation details (eg. DB naming conventions) and maintain a strict interface for us to interact with.

tgandrews: I have opened a tech debt ticket to discuss. I don't think we need classes or anything like that just make the repository boundary stricter. https://trello.com/c/shtp5ndd/331-repository-api-is-leaky

PR #126, opened 2018-09-18
Add default metadata

adammahmood:
Adds default values when updating Metadata

### How to review 
- Run tests
- Ensure default values are populate as expected

### DIFF
--- a/repositories/MetadataRepository.js
+++ b/repositories/MetadataRepository.js
@@ -1,13 +1,13 @@
 const { head } = require("lodash/fp");
+const { find, merge } = require("lodash");
 
 const Metadata = require("../db/Metadata");
 
-const TYPE_VALUE = {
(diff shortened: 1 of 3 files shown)


adammahmood:
@samiwel Struggling to reproduce the database constraint violation.  The default value should not be causing this:

```
{
    key: "country",
    alias: "Country",
    type: "TEXT",
    value: "E"
  }
```

adammahmood:
@samiwel The DB error should be fixed now.  Fixed by https://github.com/ONSdigital/eq-author-api/pull/126/commits/7f071830fe18ffdcc20a3796029e624e9c5e9285

samiwel:
@adammahmood Your latest change to fix the capitalisation of the metadata type from `TEXT` to `Text` seems to have resolved the DB constraint violation for me 👍

PR #127, opened 2018-09-19
Changing the default latest date config to match designs

SamGodwin2:
There was a small issue where the default values for the config on latest date didn't quite match the designs. This Pr brings them back into line with one another.

### How to review 
Tests should pass and the migration should successfully change all latest date configs to the correct value.


SamGodwin2:
No longer needed due to https://github.com/ONSdigital/eq-author-api/pull/128

PR #131, opened 2018-09-20
Add alias to QuestionPage

adammahmood:
- Adds `alias` field to `QuestionPage`
- https://trello.com/c/Zf8oFS8D/335-aliases-for-questions-6

### How to review 
- Run tests

```
mutation CreateQuestionPage($input: CreateQuestionPageInput!) {
  createQuestionPage(input: $input) {
    alias
  }
}

#Query Variables
{
  "input": {
    "title": "fooBarTitle",
    "alias": "fooBarAlias",
    "sectionId": "1"
  }
}
```

```
mutation UpdateQuestionPage($input: UpdateQuestionPageInput!) {
  updateQuestionPage(input: $input) {
    alias
  }
}

#Query Variables
{
  "input": {
    "id": "1",
    "alias": "fooBarAlias"
  }
}
```

```
query GetQuestionPage($id: ID!) {
  questionPage(id: $id) {
    title
    displayName
    alias
  }
}

#Query Variables
{
  "id": "1"
}
```

### Dependencies 

- [x] https://github.com/ONSdigital/eq-author-graphql-schema/pull/67


adammahmood:
[Backend changes for aliases](https://trello.com/c/5pdOqrqA/612-backend-changes-for-aliases)

repositories/PageRepository.test.js
```diff
@@ -494,13 +496,33 @@ describe("PagesRepository", () => {
...
       );
     });
 
+    it("prepends 'Copy of' to the question alias of the copy when duplicating a page", async () => {
+      const { section } = await setup();
+
+      const page = buildPage({ sectionId: section.id });
+      const result = await PageRepository.insert(page);
+
+      const duplicatePage = await PageRepository.duplicatePage(result.id, 1);
```
tgandrews: Isn't this all handled in the DuplicateStrategy?

tgandrews: Ah, I see it calls it internally. But surely it should handle the name thing as well?

adammahmood: What name thing?

This test asserts `alias` is appended with `Copy of`.

There is a test below which as asserts `title` is appended with `Copy of`.

tgandrews: I'm questioning - why is it happening here. AFAIK the Strategy should handle that for you

adammahmood: I agree. This implementation detail should *really* be within `duplicatePageStrategy` as this method is so specific and has no other use case.  Therefore, should handle everything about duplicating a page.  

Do we want to do this in this PR, a separate PR or maybe when we tackle this refactor during the duplicate Sections feature?

tgandrews: I think the logic is already in the Strategy - so you could call it with just a  page and it might do it for you.

adammahmood: As far as I can tell, there is no logic to handle prefixing in the strategy.

PR #134, opened 2018-09-24
622 remove dependency on go launch a survey

SamGodwin2:
Currently in order to support piping of metadata a change needed to be made to survey-launcher in order to provide metadata to Runner. This PR instead removes the dependencey on survey-launcher and instead uses a new endpoint on the api in order to generate a JWT that can be passed to Runner which will include all the metadata values.  

The new endpoint: 
```
query getquestionnaires($id: ID!) {
  launch(id: $id){
    location
  }
}
```

In order to support metadata however a change had to be made to remove certain metadata keys from being included in the JWT as these would have the potential to break the token. These inappropriate keys also had their defaults removed to help prevent them from being used.

 
### How to review 
Tests pass and the new endpoint can be quired and returns a valid JWT that can be passed to Runner.


docker-compose.yml
```diff
@@ -9,13 +9,15 @@ services:
       - db
     volumes:
       - .:/app
-      #- ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
+      - ./node_modules/eq-author-graphql-schema:/app/node_modules/eq-author-graphql-schema
```
tgandrews: I presume this is a mistake

SamGodwin2:
[Remove dependency on go-launch-a-survey](https://trello.com/c/FJFyZVEK/622-remove-dependency-on-go-launch-a-survey)

utils/jwtGeneration/jwtHelper.js
```diff
@@ -0,0 +1,77 @@
+/* eslint-disable camelcase*/
+const KJUR = require("jsrsasign");
+const fs = require("fs");
+const JSONWebKey = require("json-web-key");
+const jose = require("node-jose");
+const JWK = jose.JWK;
```
tgandrews: Could 5 - 7 be `const { JWK, JWE } = require("node-jose");`?

utils/jwtGeneration/jwtHelper.js
```diff
@@ -0,0 +1,77 @@
...
+  fs.readFileSync("./utils/jwtGeneration/keys.yml", "utf8")
+);
+const keysJson = JSON.parse(JSON.stringify(keysYaml, null, 4));
+
+module.exports.generateToken = function(questionnaireId, metadata) {
+  const signingKeyObject = getKeyByUse(keysJson.keys, "signing");
+  const encryptionKeyObject = getKeyByUse(keysJson.keys, "encryption");
+
+  // Header
+  const oHeader = {
```
tgandrews: What is the `o` prefix? Is it meant to signify an Object?

utils/jwtGeneration/jwtHelper.js
```diff
@@ -0,0 +1,77 @@
...
+    alg: "RS256",
```
tgandrews: Could we lift this and the value below as constants with nicer names. Especially as `RS256` is reused below.

utils/refineMetadata.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable camelcase*/
+const { omit, assign } = require("lodash/fp");
+const uuid = require("uuid/v1");
+const KJUR = require("jsrsasign");
```
tgandrews: `KJUR`? I don't know this acronym

utils/refineMetadata.js
```diff
@@ -0,0 +1,36 @@
...
+  ru_name: "ESSENTIAL ENTERPRISE LTD",
+  trad_as: "ESSENTIAL ENTERPRISE LTD",
+  eq_id: questionnaireId,
+  collection_exercise_sid: uuid(),
+  period_id: "201605",
+  form_type: questionnaireId,
+  survey_url: `${process.env.PUBLISHER_URL}${questionnaireId}`
+});
+
+module.exports.refineMetadata = (metadata, questionnaireId) => {
```
tgandrews: Could we call this "sanitise" or something similar - I'm not sure refine is the best verb to use.

utils/jwtGeneration/jwtHelper.js
SamGodwin2: I think so... It's the variable name the guy who wrote the module uses in his tutorial so I kinda just kept it. Maybe its worth changing that and the KJUR names to something a little clearer.

repositories/LauncherRepository.js
```diff
@@ -0,0 +1,13 @@
+const { generateToken } = require("../utils/jwtGeneration/jwtHelper");
+const { assign } = require("lodash");
+
+module.exports.launchASurvey = async function(id, ctx) {
```
samiwel: I would drop the `A` in this name.

utils/jwtGeneration/jwtHelper.js
```diff
@@ -0,0 +1,77 @@
...
+
+const getKeyByUse = (json, useCase) =>
+  flow(
+    keys,
+    map(kid => assign(json[kid], { kid })),
+    find(keyObject => keyObject.use === useCase)
+  )(json);
+
+const keysYaml = yaml.safeLoad(
+  fs.readFileSync("./utils/jwtGeneration/keys.yml", "utf8")
```
samiwel: I think putting this file in the project root will make things easier when we eventually grab this file out of S3.

samiwel: Also, I think we should introduce another environment variable here too. Runner call it something like `KEYS_FILE`. We should do the same even if we just default to this for now.

utils/jwtGeneration/jwtHelper.js
```diff
@@ -0,0 +1,77 @@
...
+  // Payload
+  const oPayload = refineMetadata(metadata, questionnaireId);
+
+  // Sign JWT, password=616161
+  const sHeader = JSON.stringify(oHeader);
+  const sPayload = JSON.stringify(oPayload);
+
+  const prvKey = KJUR.KEYUTIL.getKey(signingKeyObject.value, "digitaleq");
+
+  const sJWT = KJUR.jws.JWS.sign("RS256", sHeader, sPayload, prvKey);
```
samiwel: I think we can do better with the naming of variables here. `signedJWT`, `signedPayload`, `signingKey` etc. Crypto can be confusing enough we should make our lives easier where we can.

utils/jwtGeneration/jwtHelper.js
```diff
@@ -0,0 +1,77 @@
...
+      const token =
+        result.protected +
+        "." +
+        result.recipients[0].encrypted_key +
+        "." +
+        result.iv +
+        "." +
+        result.ciphertext +
+        "." +
+        result.tag;
```
samiwel: I think we can avoid string concatenation here and use JS template string.

utils/jwtGeneration/jwtHelper.js
```diff
@@ -0,0 +1,77 @@
...
+  const oHeader = {
+    alg: "RS256",
+    typ: "JWT",
+    kid: signingKeyObject.kid
+  };
+
+  // Payload
+  const oPayload = refineMetadata(metadata, questionnaireId);
+
+  // Sign JWT, password=616161
```
samiwel: What is this password in the comment? We should remove.

utils/jwtGeneration/jwtHelper.js
```diff
@@ -0,0 +1,77 @@
...
+  const sHeader = JSON.stringify(oHeader);
+  const sPayload = JSON.stringify(oPayload);
+
+  const prvKey = KJUR.KEYUTIL.getKey(signingKeyObject.value, "digitaleq");
```
samiwel: We should probably extract the passcode into an environment variable.

utils/jwtGeneration/keys.yml
```diff
@@ -0,0 +1,54 @@
...
+      MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAt8LZnIhuOdL/BC029GOa
+      JkVUAqgp2PcmbFr2Qwhf/514DUUQ9sKJ1rvwvbmmW2zE8JRtdY3ey0RXGtMn5UZH
+      s8NReHzMxvsmHN4VuaGEnFmPwO821Tkvg0LpKsLkotcw793FD/fut44N2lhpTSW2
+      Sc82uG0p9A+Kud8HCIaWaluosghk9rbMGYDzZQk8cA91GtKJRmIOED4PorB/dexD
+      f37qhuWNQgzyNyTti1DTDUIWyzQQJp926vLbkOip6Fc2R13hOFNETe68Rrw/h3hX
+      EFS17uPFZHsxvm9PFXX9KZMS25ohqbNh97I94LL4o4wybl6LaE6lJEHiD6docD0B
+      6wIDAQAB
+      -----END PUBLIC KEY-----
+
+    version: sr
```
samiwel: As a separate note. It would be good to document how we generated this file somewhere in confluence if it doesn't already exist.

SamGodwin2: I doubt it exists, I'll take a look and if not write something up.

utils/sanitiseMetadata.test.js
```diff
@@ -0,0 +1,56 @@
...
+      user_id: "UNKNOWN",
+      case_id: 123,
+      ru_ref: "12346789012A",
+      ru_name: "ESSENTIAL ENTERPRISE LTD",
+      trad_as: "ESSENTIAL ENTERPRISE LTD",
+      eq_id: 1,
+      collection_exercise_sid: 123,
+      period_id: "201605",
+      form_type: 1,
+      survey_url: `${process.env.PUBLISHER_URL}1`
```
samiwel: Do we need to set the env var for the test? Seems like we should be able to put whatever here?

utils/sanitiseMetadata.js
```diff
@@ -0,0 +1,42 @@
...
+  case_id: uuid(),
+  ru_ref: "12346789012A",
+  ru_name: "ESSENTIAL ENTERPRISE LTD",
+  trad_as: "ESSENTIAL ENTERPRISE LTD",
+  eq_id: questionnaireId,
+  collection_exercise_sid: uuid(),
+  period_id: "201605",
+  form_type: questionnaireId,
+  survey_url: `${process.env.PUBLISHER_URL}${questionnaireId}`
+});
```
samiwel: Can the survey_url be passed in from above?

utils/jwtHelper.js
```diff
@@ -0,0 +1,81 @@
...
+  flow(
+    keys,
+    map(kid => assign(json[kid], { kid })),
+    find(keyObject => keyObject.use === useCase)
+  )(json);
+
+const keysYaml = yaml.safeLoad(
+  fs.readFileSync(`${process.env.KEYS_FILE}`, "utf8")
+);
+const keysJson = JSON.parse(JSON.stringify(keysYaml, null, 4));
```
samiwel: No benefit to pretty printing here?

utils/sanitiseMetadata.js
```diff
@@ -0,0 +1,43 @@
...
+  "eq_id",
+  "form_type",
+  "account_service_url"
+]);
+
+const defaultMetadata = (questionnaireId, tokenIssueTime, surveyUrl) => ({
+  tx_id: uuid(),
+  jti: uuid(),
+  iat: tokenIssueTime,
+  exp: tokenIssueTime + 10000,
```
tgandrews: Can we extract this `10000` as a constant - I presume its the expiry time but in what unit. Is it 10000 seconds or milliseconds?

utils/jwtHelper.js
```diff
@@ -0,0 +1,77 @@
...
+  // Header
+  const jwtHeader = JSON.stringify({
+    alg: SIGNING_ALGORITHM,
+    typ: "JWT",
+    kid: signingKeyObject.kid
+  });
+
+  const tokenIssuedTime = KJUR.jws.IntDate.get("now");
+
+  const payload = sanitiseMetadata(
```
tgandrews: Could we pass this all down to this function instead of it knowing anything about metadata sanitising. It feels like the rest of the function is token specific and this payload generation isn't. So we would call sanitiseMetadata in the launch middleware

tgandrews:
Could we update the PR description to reflect the design change.

utils/jwtHelper.js
SamGodwin2: I did have it set-up like that originally however it meant we needed to import the KJUR module both here for the crypto stuff and then also in the launch middleware for the expiry time. I moved it here to try to isolate the crypto imports to one module but understand it makes this function a little less specific. Not sure which approach would be better here?

middleware/launch.test.js
```diff
@@ -0,0 +1,49 @@
+const getLaunchUrl = require("./launch");
+const mockRepository = require("../tests/utils/mockRepository");
+
+let repositories;
+const QUESTIONNAIRE_ID = 1;
+
+describe("status middleware", () => {
```
samiwel: copy and paste error

middleware/launch.test.js
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  it("should call a redirect with a jwt", async () => {
+    const ctx = { repositories };
+    await getLaunchUrl(ctx)(req, res, next);
+
+    expect(res.redirect).toHaveBeenCalled();
+  });
+
+  it("should call next with an error if a metadata is missing a key", async () => {
```
samiwel: Probably also worth adding tests describing the behaviour of `null` and "    ". Since keys are null by default?

utils/sanitiseMetadata.js
```diff
@@ -0,0 +1,41 @@
...
+  user_id: "UNKNOWN",
+  case_id: uuid(),
+  ru_ref: "12346789012A",
+  ru_name: "ESSENTIAL ENTERPRISE LTD",
+  trad_as: "ESSENTIAL ENTERPRISE LTD",
+  eq_id: questionnaireId,
+  collection_exercise_sid: uuid(),
+  period_id: "201605",
+  form_type: questionnaireId,
+  survey_url: `${surveyUrl}${questionnaireId}?r${Date.now()}`
```
samiwel: Wonder if `tokenIssueTime` can be used instead of `Date.now()`. Do they produce the same format of output?

keys.yml
```diff
@@ -0,0 +1,54 @@
...
+      pDAlNMCA95hN5/8LFIYsUXL/nCbgY0gsd8K5po9ekZCCnpTh1sr61h9jk24mZUz6
+      uyyq94IrWfIGqSfi4DF/42LKdrPm8kU5DNRR4ZOaU3aQpKMt84KyQXL7ElyDLyPD
+      yj5Hm9xF+6mSPYzJJAItYQKBgHzUZXbzf7ZfK2fwVSAlt68BJDvnzP62Z95Hqgbp
+      hjDThXPbvBXYcGkt1fYzIPZPeOxe6nZv/qGOcEGou4X9nOogpMdC09qprTqw/q/N
+      w9vUI3SaW/jPuzeqZH7Mx1Ajhh8uC/fquK7eMe2Dbi0b2XOeB08atrLyhk3ZEMsL
+      2+IFAoGAUbmo0idyszcarBPPsiEFQY2y1yzHMajs8OkjUzOVLzdiMkr36LF4ojgw
+      UCM9sT0g1i+eTfTcuOEr3dAxcXld8Ffs6INSIplvRMWH1m7wgXMRpPCy74OuxlDQ
+      xwPp/1IVvrMqVgnyS9ezAeE0p9u8zUdZdwHz1UAggwbtHR6IbIA=
+      -----END RSA PRIVATE KEY-----
+    version: rrm
```
samiwel: Apparently this version is supposed to be numeric so that we can rotate the keys e.g. v1, v2, v3 etc. I don't think it matters a great deal for now. But just for info.

utils/jwtHelper.js
```diff
@@ -0,0 +1,68 @@
+/* eslint-disable camelcase*/
+const KJUR = require("jsrsasign");
+const fs = require("fs");
+const JSONWebKey = require("json-web-key");
+const { JWK, JWE } = require("node-jose");
+const { find, keys, assign, flow, map } = require("lodash/fp");
+const yaml = require("js-yaml");
+const SIGNING_ALGORITHM = "RS256";
+const keysFile = process.env.KEYS_FILE || "./keys.yml";
+const signingPassword = process.env.SIGNING_PASSWORD || "digitaleq";
```
samiwel: As discussed, this can be removed 👍

PR #137, opened 2018-09-25
Add displayName for metadata

tgandrews:
Adds a displayName for metadata so we don't have to do this in the UI.

### How to review 
1. Ensure tests pass.
2. Ensure that metadata can be retrieved with a displayName

### Dependencies
- [x] - https://github.com/ONSdigital/eq-author-graphql-schema/pull/69

### DIFF
--- a/utils/getName.test.js
+++ b/utils/getName.test.js
@@ -61,4 +61,15 @@ describe("getName", () => {
       expect(getName(entity, typeName)).toEqual(entity.label)
     );
   });
+
+  it("should use key if none of the other fields are avaialable", () => {
+    entity = {
+      foo: "I am a title",
+      key: "I am a key"
+    };
+
+    map(keys(defaultNames), typeName =>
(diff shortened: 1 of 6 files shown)


utils/getName.test.js
```diff
@@ -61,4 +61,15 @@ describe("getName", () => {
       expect(getName(entity, typeName)).toEqual(entity.label)
     );
   });
+
+  it("should use key if none of the other fields are avaialable", () => {
```
adammahmood: Should this not be part of the `should use correct key` test?  Also, the wording for this scenario will become redundant if we add another fallback key.

tgandrews: `should use correct key` doesn't test the precedence - just the field to use first.

tgandrews:
[Pipe metadata into a question - front end changes](https://trello.com/c/fkHnQhid/536-pipe-metadata-into-a-question-front-end-changes)

PR #138, opened 2018-09-26
Remove deprecated fields

rmccar:
Closes #112

Remove deprecated fields

### How to review 
Make sure tests pass, run system to make sure it runs without error


tests/schema/mutations/updateAnswer.test.js
```diff
@@ -45,25 +44,4 @@ describe("updateAnswer", () => {
     expect(result.errors).toBeUndefined();
     expect(repositories.Answer.update).toHaveBeenCalledWith(input);
   });
-
-  it("should allow update of Answer with deprecated fields", async () => {
```
tgandrews: Are we not allowing this anymore?

PR #149, opened 2018-10-04
Modify docker config to read secrets from S3 bucket

samiwel:

A change is required to the author API so that it is possible to read in secrets from an S3 bucket at runtime.

This is required because the signing and encryption keys used to create the JWT payload differ between dev, preprod and prod. The new launch mechanism works in staging at the moment due to the use of development keys, but will not currently work in pre-prod because the keys are not the same.

This change allows for overriding of the default behaviour with a new environment variable. This allows secrets to be read in from S3 at runtime.

I've also made a few optimisations to how we build the image to speed things up (e.g. no longer copying the contents of node_modules into the container) and updated the README with a few missing environment variable details.

### How to review 
 - Build the docker image using `docker build -t eq-author-api:8-alpine-with-awscli .`
 - Run the container using `docker run -ti eq-author-api:8-alpine-with-awscli`
   The container will fail to run due to missing database.
 - Kill the container and run again with `docker run -ti -e SECRETS_S3_BUCKET=env-secrets eq-author-api:8-alpine-with-awscli`
 - Observe in the output that an attempt was made to read secrets in from env-secrets.

As a final check it's also worth running `docker-compose up --build` to make sure the developer experience is the same/similar to before.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,24 @@
 FROM node:8-alpine
+
+RUN apk add --no-cache python3 && \
+    python3 -m ensurepip && \
+    rm -r /usr/lib/python*/ensurepip && \
+    pip3 install --upgrade pip setuptools && \
+    if [ ! -e /usr/bin/pip ]; then ln -s pip3 /usr/bin/pip ; fi && \
+    if [ ! -e /usr/bin/python ]; then ln -sf /usr/bin/python3 /usr/bin/python; fi && \
+    rm -r /root/.cache && \
+    pip install --upgrade awscli==1.16.27
+
 EXPOSE 4000
 ENV PORT=4000
 WORKDIR /app
 
 ARG APPLICATION_VERSION
 ENV EQ_AUTHOR_API_VERSION $APPLICATION_VERSION
+ENV AWS_DEFAULT_REGION eu-west-1
 ENV NODE_ENV production
 
-ENTRYPOINT ["yarn", "start"]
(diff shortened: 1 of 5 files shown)


samiwel:
[Read encryption and signing keys from the environment](https://trello.com/c/4LGWQyFO/653-read-encryption-and-signing-keys-from-the-environment)

Dockerfile
```diff
@@ -1,13 +1,18 @@
 FROM node:8-alpine
+
+COPY . /app
```
jonnyshaw89: Why move this copy to the top? it will make layer re-use really bad.

samiwel: Hmm, good point. I didn't consider that.

samiwel: Yeah I thought about doing that that but inlining isn't so bad. At least it's all in one place. It's not like we're going to re-use the script anywhere else.

Dockerfile
```diff
@@ -1,13 +1,24 @@
...
 ENV PORT=4000
 WORKDIR /app
 
 ARG APPLICATION_VERSION
 ENV EQ_AUTHOR_API_VERSION $APPLICATION_VERSION
+ENV AWS_DEFAULT_REGION eu-west-1
 ENV NODE_ENV production
 
-ENTRYPOINT ["yarn", "start"]
+ENTRYPOINT ["sh", "docker-entrypoint.sh"]
```
tgandrews: I don't think the `"sh"` is needed

PR #152, opened 2018-10-12
Add previous answer to max value

adammahmood:
- Adds validation against a previous answer for max value 

### How to review 
- Run tests

```
mutation updateValidationRule($input: UpdateValidationRuleInput!) {
  updateValidationRule(input: $input) {
    ...MaxValueValidationRule
    __typename
  }
}

fragment MaxValueValidationRule on MaxValueValidationRule {
  id
  enabled
  custom
  inclusive
  entityType
  previousAnswer {
    id
    displayName
    __typename
  }
  __typename
}


#Query Variables
{
  "input": {
    "id": "2", 
    "maxValueInput": {
      "inclusive": false, 
      "previousAnswer": "1"
    }
  }
}

```

```
query answerWithValidation($id: ID!) {
  answer(id: $id) {
  	... on BasicAnswer {
      validation {
        ... on NumberValidation {
         	maxValue {
            id
            previousAnswer {
              id
              displayName
            }
          } 
        }
      }
    }
  }
}

#Query Variables
{
  "id": "2"
}
```

### Dependencies 

- [x] https://github.com/ONSdigital/eq-author-graphql-schema/pull/77

### DIFF
--- a/schema/resolversTests/validation.test.js
+++ b/schema/resolversTests/validation.test.js
@@ -154,17 +154,18 @@ describe("resolvers", () => {
           id: maxValueId,
           enabled: false,
           inclusive: false,
-          custom: null
+          custom: null,
+          entityType: "Custom"
         }
       });
 
-      expect(currencyValidation).toMatchObject(
+      expect(currencyValidation).toEqual(
         validationObject(
           currencyValidation.minValue.id,
           currencyValidation.maxValue.id
         )
       );
-      expect(numberValidation).toMatchObject(
+      expect(numberValidation).toEqual(
         validationObject(
           numberValidation.minValue.id,
           numberValidation.maxValue.id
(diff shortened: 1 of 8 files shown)


schema/resolversTests/validation.test.js
```diff
@@ -211,6 +211,30 @@ describe("resolvers", () => {
...
     });
+
+    it("can update inclusive and previous answer max values", async () => {
+      const previousAnswer = await createNewAnswer(firstPage, "Number");
+      const currencyAnswer = await createNewAnswer(firstPage, "Currency");
+      const currencyValidation = await queryAnswerValidations(
+        currencyAnswer.id
+      );
+
+      const result = await mutateValidationParameters({
```
tgandrews: Does `entityType` not need to be passed?

adammahmood: Just added separate tests for entityType.  It is updated independently.

schema/resolversTests/validation.test.js
```diff
@@ -205,12 +206,61 @@ describe("resolvers", () => {
...
+
+    it("can update inclusive and entity type", async () => {
+      const currencyAnswer = await createNewAnswer(firstPage, "Currency");
+      const currencyValidation = await queryAnswerValidations(
+        currencyAnswer.id
+      );
+
+      const entityTypes = ["Custom", "PreviousAnswer"];
+
+      entityTypes.map(entityType => {
```
tgandrews: Should this be `forEach` and `await`ing each mutation?

tgandrews: Also, it doesn't return the promise array - so would this fail?

adammahmood: This works fine because `map` is synchronous.  Each request is resolved within the map so no need to return a promise array.

PR #154, opened 2018-10-16
656 api changes for routing based on number

SamGodwin2:
Api backend changes to implement routing based on a number.

### How to review 
Tests pass and the new functionality works.

### DIFF
--- a/migrations/20181004135102_add_custom_values_to_routing_condition_values.js
+++ b/migrations/20181004135102_add_custom_values_to_routing_condition_values.js
@@ -1,0 +1,20 @@
+const formatAlterTableEnumSql = require("../utils/migrateEnumChecks");
+const { noop } = require("lodash");
+
+exports.up = async function(knex) {
+  await knex.raw(
+    formatAlterTableEnumSql("Routing_Conditions", "comparator", [
+      "Equal",
+      "NotEqual",
+      "GreaterThan",
+      "LessThan",
+      "GreaterOrEqual",
+      "LessOrEqual"
+    ])
+  );
+  return knex.schema.table("Routing_ConditionValues", table => {
+    table.integer("customNumber").defaultsTo(null);
+  });
+};
+
+exports.down = noop;
--- a/repositories/RoutingRepository.js
+++ b/repositories/RoutingRepository.js
@@ -1,6 +1,7 @@
-const { head, map } = require("lodash/fp");
+const { head } = require("lodash/fp");
 const { get, isNil, parseInt } = require("lodash");
 const db = require("../db");
+const Answer = require("../repositories/AnswerRepository");
 
 const {
   updateRoutingConditionStrategy,
@@ -81,9 +82,7 @@ function findAllRoutingConditions(where = {}) {
 }
 
 function findAllRoutingConditionValues(where = {}) {
-  return Routing.findAllRoutingConditionValues()
-    .where(where)
-    .then(map("optionId"));
+  return Routing.findAllRoutingConditionValues().where(where);
 }
 
 function createRoutingRuleSet({ questionPageId }) {
@@ -118,6 +117,15 @@ const toggleConditionOption = async ({ conditionId, optionId, checked }) =>
     })
   );
 
+const createConditionValue = async ({ conditionId }) =>
+  Routing.createRoutingConditionValue({
+    conditionId,
+    customNumber: null
+  }).then(head);
+
+const updateConditionValue = async ({ id, customNumber }) =>
+  Routing.updateRoutingConditionValue(id, { customNumber }).then(head);
+
 const updateDestination = async (id, destination) => {
   const { logicalDestination, absoluteDestination } = destination;
 
@@ -150,6 +158,16 @@ async function updateRoutingRuleSet({ id, else: destination }) {
   return routingRuleSet;
 }
 
+async function getAnswerTypeByConditionId(id) {
+  const { answerId } = await findAllRoutingConditions({ id }).then(head);
+  if (isNil(answerId)) {
+    return null;
+  } else {
+    const { type } = await Answer.getById(answerId);
+    return type;
+  }
+}
+
 async function updateRoutingRule({ id, goto: destination }) {
   const routingRule = await getRoutingRuleById(id);
   const routingRuleSet = await getRoutingRuleSetById(
@@ -165,9 +183,14 @@ async function updateRoutingRule({ id, goto: destination }) {
   return routingRule;
 }
 
-function updateRoutingCondition({ id, questionPageId, answerId }) {
+function updateRoutingCondition({ id, questionPageId, answerId, comparator }) {
   return db.transaction(trx =>
-    updateRoutingConditionStrategy(trx, { id, questionPageId, answerId })
+    updateRoutingConditionStrategy(trx, {
+      id,
+      questionPageId,
+      answerId,
+      comparator
+    })
   );
 }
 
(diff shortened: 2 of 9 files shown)


migrations/20181004135102_add_custom_values_to_routing_condition_values.js
```diff
@@ -0,0 +1,27 @@
+const formatAlterTableEnumSql = (tableName, columnName, enums) => {
```
tgandrews: We should extract this to a utility.

repositories/strategies/routingStrategy.js
```diff
@@ -21,7 +28,7 @@ const getFirstAnswer = (trx, questionPageId) =>
   trx("Answers")
     .where({ questionPageId, isDeleted: false })
     .orderBy("id")
-    .first();
+    .then(head);
```
tgandrews: Why not `first`? It limits the data in the db.

repositories/strategies/routingStrategy.js
```diff
@@ -111,15 +134,17 @@ const checkAnswerBelongsToPage = async (trx, answerId, questionPageId) => {
 
 const getAnswerOrFirstAnswerOnPage = async (trx, answerId, questionPageId) => {
```
tgandrews: This doesn't need to be `async` AFAICT it never `await`s.

repositories/strategies/routingStrategy.js
```diff
@@ -206,15 +239,31 @@ const checkRoutingDestinations = async (
...
-    ...routingCondition,
-    answerId: get(firstAnswerOnPage, "id", answerId)
-  });
+  const targetAnswerId = get(firstAnswerOnPage, "id", answerId);
+
+  if (!isNil(targetAnswerId)) {
+    targetAnswer = await trx("Answers")
+      .select()
+      .where({ id: targetAnswerId })
+      .then(head);
```
tgandrews: `first()`?

repositories/strategies/routingStrategy.js
```diff
@@ -206,15 +239,31 @@ const checkRoutingDestinations = async (
 
 async function createRoutingConditionStrategy(trx, routingCondition) {
   const { questionPageId, answerId } = routingCondition;
+  let targetAnswer;
```
tgandrews: Minor nit: You don't need to hoist this. I would put it above line `250`.

SamGodwin2:
[API changes for routing based on number](https://trello.com/c/1vVcL4PS/656-api-changes-for-routing-based-on-number)

schema/resolvers.js
```diff
@@ -15,6 +15,13 @@ const assertMultipleChoiceAnswer = answer => {
   }
 };
 
+const getRoutingConditionsFields = async (conditionId, field, ctx) => {
+  const conditionValues = await ctx.repositories.Routing.findAllRoutingConditionValues(
+    { conditionId }
+  );
+  return first(conditionValues.map(conditionValue => conditionValue[field]));
```
tgandrews: Why are we mapping the whole array - can't we just pull off the field from the first result.

`conditionValues[0][field]`

schema/resolvers.js
```diff
@@ -263,12 +274,35 @@ const Resolvers = {
...
 
   RoutingConditionValue: {
-    __resolveType: () => "IDArrayValue"
+    __resolveType: async ({ conditionId }, ctx) => {
+      const answerType = await ctx.repositories.Routing.getAnswerTypeByConditionId(
+        conditionId,
+        ctx
+      );
+      if (answerType === "Radio") {
+        return "IDArrayValue";
```
tgandrews: Thats a very funny type name. But existing - just commenting.

schema/resolvers.js
```diff
@@ -249,8 +260,8 @@ const Resolvers = {
   },
 
   RoutingCondition: {
-    routingValue: ({ id }) => {
-      return { conditionId: id };
+    routingValue: ({ id, answerId }) => {
+      return { conditionId: id, answerId };
```
tgandrews: Could we return the whole condition here? So that the field resolvers would just be picking them off the existing object?

schema/resolvers.js
```diff
@@ -263,12 +274,35 @@ const Resolvers = {
...
+        {
+          conditionId
+        }
+      );
+      return conditionValues.map(conditionValue => conditionValue.optionId);
+    }
+  },
+
+  NumberValue: {
+    id: async ({ conditionId }, args, ctx) =>
```
tgandrews: This looks odd - we are making a query for each field instead of picking off an already exiting entity. Maybe I'm missing something.

schema/resolvers.test.js
```diff
@@ -1461,5 +1465,81 @@ describe("resolvers", () => {
...
+        const answer = await createNewAnswer(firstPage, "Currency");
+        await createNewRoutingRuleSet(firstPage.id);
+
+        const routingInfo = await getFullRoutingTree(firstPage);
+        const routingCondition = get(
+          routingInfo,
+          "data.questionPage.routingRuleSet.routingRules[0].conditions[0]"
+        );
+
+        expect(routingCondition).toMatchObject({
```
tgandrews: These are default values? I can't see where they are defined in this PR.

tests/utils/graphql.js
```diff
@@ -235,9 +235,14 @@ query GetPage($id: ID!){
           conditions{
             id
             routingValue{
+
```
tgandrews: Nit: No need for space

repositories/RoutingRepository.js
```diff
@@ -150,6 +157,16 @@ async function updateRoutingRuleSet({ id, else: destination }) {
   return routingRuleSet;
 }
 
+async function getAnswerTypeByConditionId(id, ctx) {
```
tgandrews: I'm not sure I like the repository getting the `ctx`. Can't we just import the other repository?

schema/resolvers.test.js
SamGodwin2: Equal is the default value, it would have been defined in the first implementation of routing, this one just adds other comparators without changing the default.

migrations/20181004135102_add_custom_values_to_routing_condition_values.js
```diff
@@ -0,0 +1,18 @@
...
+      "GreaterOrEqual",
+      "LessOrEqual"
+    ])
+  );
+  return knex.schema.table("Routing_ConditionValues", table => {
+    table.integer("customNumber").defaultsTo(null);
+  });
+};
+
+exports.down = function() {};
```
adammahmood: Do we even need to implement this method? Can it be removed entirely?

repositories/strategies/routingStrategy.js
```diff
@@ -67,12 +74,28 @@ const getRoutingRuleSetById = (trx, routingRuleSetId) => {
...
+    .returning("*")
+    .then(head);
+
+const insertRoutingCondition = async (trx, routingCondition, answer) => {
+  const condition = await trx("Routing_Conditions")
     .insert(routingCondition)
     .returning("*")
     .then(head);
 
+  if (!isNil(answer) && includes(["Currency", "Number"], answer.type)) {
```
adammahmood: Although, we don't currently have this and it's beyond the scope of this task.  I think at some point in the future we should make the answer types constants.

migrations/20181004135102_add_custom_values_to_routing_condition_values.js
SamGodwin2: Annoyingly not, if we try remove this knex throws an error.

samiwel: Lodash has a `noop` function that may help convey semantically that `exports.down = noop`

PR #158, opened 2018-10-17
Fix issue piping Date metadata.

samiwel:
It is not currently possible to pipe Date metadata into a questionnaire. Survey runner expects date values to be passed through in ISO format (yyyy-mm-dd) and author currently passes them in dd/mm/yyyy format.

The result is that the survey will launch but runner will throw a 403 authentication error (since the JWT token is considered invalid).

This fix formats Date values according to ISO format before they are passed into the claims of the JWT token. After applying this change, it should be possible to launch a questionnaire that pipes in a Date metadata value and have it show up within runner.

### How to review 
- Run the application using eq-compose
- Confirm that you cannot launch a questionnaire that pipes a Date metadata value
- Change the config of eq-compose so that it pulls down this branch's image for the Author API.
- Run the application again and confirm you can pipe Date metadata values successfully and launch into runner showing the value.
- All existing checks and tests should pass.

### DIFF
--- a/middleware/launch.js
+++ b/middleware/launch.js
@@ -2,21 +2,36 @@ const { generateToken } = require("../utils/jwtHelper");
 const { assign, isNil, isEmpty } = require("lodash");
 const { sanitiseMetadata } = require("../utils/sanitiseMetadata");
 
-module.exports = ctx => async (req, res, next) => {
+const buildClaims = metadata => {
+  const result = {
+    claims: {},
+    errors: []
+  };
+
+  metadata.map(({ key, value, id, type }) => {
+    if (isNil(key) || key.trim() === "") {
+      result.errors.push(id);
+    }
+
+    return assign(result.claims, {
+      [key]:
+        type === "Date" ? new Date(value).toISOString().split("T")[0] : value
+    });
+  });
+
+  return result;
+};
+
+module.exports.buildClaims = buildClaims;
+
+module.exports.getLaunchUrl = ctx => async (req, res, next) => {
   const questionnaireId = req.params.questionnaireId;
-  const errors = [];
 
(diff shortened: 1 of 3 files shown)


middleware/launch.test.js
```diff
@@ -85,4 +85,33 @@ describe("launcher middleware", () => {
 
     expect(next).toHaveBeenCalledWith(expect.any(Error));
   });
+
+  it("should convert date values to ISO string dates when building claims", () => {
+    expect(
+      buildClaims([
```
tgandrews: We are holding date in two different formats?

samiwel: It seems so. We could look at storing date values in ISO format, but then you need to convert back and forth on each query/mutation assuming that the author wants to see the date in dd/mm/yyyy format - we could confirm with Sarah). Converting at the point of launch isolates the change to a single place.

middleware/launch.js
```diff
@@ -2,21 +2,36 @@ const { generateToken } = require("../utils/jwtHelper");
...
+        type === "Date" ? new Date(value).toISOString().split("T")[0] : value
+    });
+  });
+
+  return result;
+};
+
+module.exports.buildClaims = buildClaims;
+
+module.exports.getLaunchUrl = ctx => async (req, res, next) => {
```
tgandrews: The api has changed so `app.js` fails to use getLaunchUrl on line `14`

tgandrews:
Works e2e when I made the `app.js` change locally.

samiwel:
[Bug - unable to pipe date metadata](https://trello.com/c/wsDdSjgd/687-bug-unable-to-pipe-date-metadata)

PR #161, opened 2018-10-19
679  - Duplicate piping

tgandrews:
This implements:
1. Easier setup of complex questionnaires for tests
1. Duplication of piping when duplicating questionnaires and sections

### How to review 
1. Checkout author branch `551-duplicate-questionnaire`
1. Create a questionnaire with piping
1. Duplicate a questionnaire/section
1. Ensure that the piping is updated as expected:
	1. When duplicating a section - references within a section are updated but external references are not
	1. When duplicating a questionnaire - all references are updated.

### DIFF
--- a/repositories/strategies/duplicateStrategy.js
+++ b/repositories/strategies/duplicateStrategy.js
@@ -1,8 +1,33 @@
 const { omit, head, get, isNil } = require("lodash");
+const cheerio = require("cheerio");
 const {
   getOrUpdateOrderForPageInsert,
   getOrUpdateOrderForSectionInsert
 } = require("./spacedOrderStrategy");
+
+const updatePiping = (field, references) => {
+  if (!field || field.indexOf("<span") === -1) {
+    return field;
+  }
+
+  const $ = cheerio.load(field);
+
+  $("span").map((i, el) => {
+    const $el = $(el);
+    const pipeType = $el.data("piped");
+    const id = $el.data("id");
+
+    const newId = references[pipeType][id];
+
+    // Can't use data as it doesn't work
+    // https://github.com/cheeriojs/cheerio/issues/1240
+    $el.attr("data-id", newId);
+
+    return $.html($el);
+  });
+
+  return $("body").html();
+};
 
 const insertData = async (
   trx,
@@ -118,13 +143,20 @@ const duplicateOtherAnswer = async (trx, otherAnswer, duplicateAnswer) => {
   });
 };
 
-const duplicateAnswerStrategy = async (trx, answer, overrides = {}) => {
+const duplicateAnswerStrategy = async (
+  trx,
+  answer,
+  overrides = {},
+  references
+) => {
   const duplicateAnswer = await duplicateRecord(
     trx,
     "Answers",
     answer,
     overrides
   );
+  references.answers[answer.id] = duplicateAnswer.id;
+
   const otherAnswer = await selectData(trx, "Answers", "*", {
     parentAnswerId: answer.id
   }).then(head);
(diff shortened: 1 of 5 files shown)


tgandrews:
[Duplicate piping](https://trello.com/c/z9B8Zp38/679-duplicate-piping)

repositories/strategies/duplicateStrategy.js
```diff
@@ -194,16 +240,20 @@ const duplicatePageStrategy = async (trx, page, position, overrides = {}) => {
     isDeleted: false
   });
 
-  await Promise.all(
-    answersToDuplicate.map(answer =>
-      duplicateAnswerStrategy(trx, answer, {
+  for (let i = 0; i < answersToDuplicate.length; ++i) {
+    const answer = answersToDuplicate[i];
+    await duplicateAnswerStrategy(
```
adammahmood: Do we need to `await` here?

repositories/strategies/duplicateStrategy.js
```diff
@@ -226,22 +277,37 @@ const duplicateSectionStrategy = async (
     sectionId: section.id
   });
 
-  await Promise.all(
-    pagesToDuplicate.map(({ position, ...page }) =>
-      duplicatePageStrategy(trx, page, position, {
+  for (let i = 0; i < pagesToDuplicate.length; ++i) {
+    const { position, ...page } = pagesToDuplicate[i];
+    await duplicatePageStrategy(
```
adammahmood: Do we need to await here?

repositories/strategies/buildTestQuestionnaire.js
```diff
@@ -0,0 +1,254 @@
...
+    createdBy: "foo",
+    ...questionnaireProps
+  });
+
+  const references = {
+    answers: {},
+    metadata: {}
+  };
+
+  questionnaire.metadata = await buildMetadata(
```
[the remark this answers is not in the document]
tgandrews: Sections reference metadata in piping so we need to have all metadata before creating sections.

repositories/strategies/duplicateStrategy.js
tgandrews: Pages can reference answers in previous pages so we need them to be created before we carry on.

adammahmood:
Works great e2e 👍

repositories/strategies/duplicateStrategy.js
tgandrews: We don't. Nice spot.

repositories/strategies/buildTestQuestionnaire.js
```diff
@@ -0,0 +1,254 @@
...
+
+  return field.replace(`data-id="${id}"`, `data-id="${newId}"`);
+};
+
+const buildValidations = async (validationConfigs = {}, answer) => {
+  const validationEntityType = getValidationEntity(answer.type);
+  const validationTypes = validationRuleMap[validationEntityType] || [];
+
+  let validations = {};
+  for (let i = 0; i < validationTypes.length; ++i) {
```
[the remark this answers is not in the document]
tgandrews: Unfortunately we can't as this would trigger all the duplications to run simultaneously.

We could create a chain of Promises instead but this is nicer. IMO

tgandrews:
Unfortunately we can't parallelise everything as we have dependencies on other things being completed so we don't have to do things with a second pass.

tgandrews:
I had a look in to the `async` library and I think if we needed to limit the amount of calls going then it would make sense to use it but for just running things in a chain we have `for() { await func() }` and then for running things in parallel we have `Promise.all(list.map(item => func())`.

PR #169, opened 2018-10-26
Changes to API to add section introduction

rmccar:
Made changes to the API to add the functionality to add introductions to sections

### How to review 
Review code changes
Run tests


rmccar:
[Section introduction page - api](https://trello.com/c/VzFQA2Qc/683-section-introduction-page-api)

migrations/20181023144034_created_introduction_columns.js
```diff
@@ -0,0 +1,21 @@
...
+        PARTITION BY "questionnaireId"
+        ORDER BY "order" ASC
+      ) - 1 AS "position"
+      FROM "Sections"
+      WHERE "isDeleted" = false
+    )`);
+};
+
+exports.down = async function(knex, Promise) {
+  return Promise.resolve();
```
SamGodwin2: Minor thing but, I think the current method for dealing with this down is to return lodash's noop function.

package.json
```diff
@@ -50,7 +50,7 @@
   "jest": {
     "testEnvironment": "node",
     "coverageDirectory": "./coverage/",
-    "collectCoverage": true,
+    "collectCoverage": false,
```
SamGodwin2: Does this need to change?

repositories/SectionRepository.js
```diff
@@ -99,3 +120,52 @@ module.exports.duplicateSection = function duplicateSection(id, position) {
...
+  content
+}) {
+  await Section.update(sectionId, {
+    introductionTitle: title,
+    introductionContent: content
+  });
+
+  return {
+    title,
+    content
```
SamGodwin2: Could we return the result from the database here rather than the input args, would add bit more assurance the data has entered the db correctly?

repositories/SectionRepository.js
```diff
@@ -99,3 +120,52 @@ module.exports.duplicateSection = function duplicateSection(id, position) {
     });
   });
 };
+
+module.exports.createIntroduction = function createIntroduction(
+  introductionInput
+) {
+  return db.transaction(async trx => {
+    await trx("Sections")
+      .update({ introductionTitle: null, introductionContent: null })
```
adammahmood: Why do you need to `null` this?

repositories/SectionRepository.js
```diff
@@ -99,3 +120,52 @@ module.exports.duplicateSection = function duplicateSection(id, position) {
     });
   });
 };
+
+module.exports.createIntroduction = function createIntroduction(
```
adammahmood: Do we need to create these CRUD methods? Can we not leverage the `update` method?

schema/resolvers.js
```diff
@@ -197,9 +205,24 @@ const Resolvers = {
         return position;
       }
       return ctx.repositories.Section.getPosition({ id });
+    },
+    introduction: section => {
+      const { introductionEnabled } = section;
+
+      return introductionEnabled
```
adammahmood: I think it's best to avoid ternaries over multiple lines.  Maybe this should be `if(...){...}else{...}` statement.

adammahmood: Or (by using destructing and renaming) you could change this to:

`return introductionEnabled ? {title, content} : null;`

Here's a good article on destructuring: https://medium.com/@pyrolistical/destructuring-nested-objects-9dabdd01a3b8

package.json
rmccar: I had a look at what it does and I don't think it does. This is what it says on jests documentation:

Indicates whether the coverage information should be collected while executing the test. Because this retrofits all executed files with coverage collection statements, it may significantly slow down your tests.

So ill change it back

rmccar: However I have just noticed that it is false in author?

adammahmood: It's best to remove this from this PR and maybe open a discussion / create a tech debt card to review this configuration.

rmccar: Yep ive reverted the change now and ill create a tech debt card

repositories/SectionRepository.js
adammahmood: Does it not get immediately overwritten by the query below?

tests/schema/mutations/updateSection.test.js
```diff
@@ -12,6 +12,38 @@ describe("updateSection", () => {
...
+    }
+  }`;
+
+  const undeleteSectionIntroduction = `
+  mutation UndeleteIntroduction($input: UndeleteIntroductionInput!) {
+    undeleteIntroduction(input: $input) {
+      title
+      content
+    }
+  }`;
```
SamGodwin2: Small nitpick but could you import these mutations from the `utils/graphql` file, saves them from being written out twice?

PR #170, opened 2018-10-26
700 - Duplicate performance

tgandrews:
This is an attempt to remove a lot of the complexity around how duplication was implemented and also change the architecture to improve speed.

This is done by instead of working through the tree entity by entity (and waiting for each entity to be completed) we now work top down through the hierarchy so we create all parents before we create any children. We also parallelise the creation of parents when they share no relationships. All of this is not done in the `duplicateTree` function in `duplicateStrategy/utils.js`. It takes a defined hierarchy and works through.

Our structure is nearly perfectly compatible with this way of working excluding piping (where pages reference answers) and routingDestinations which themselves have no reference to the parent. So in both these cases we have to query those objects again to work out what to update or duplicate.

This change took duplicating UKIS:
Before (~4 seconds):
screenshot 2018-10-26 at 15 52 50

After (~400 ms):
screenshot 2018-10-26 at 15 54 47

### How to review 
Describe the steps required to test the changes (include screenshots if appropriate).


adammahmood:
@tgandrews This now supports metadata, previous answer and custom validations.

tgandrews:
[Duplicate - Improve performance](https://trello.com/c/64sJix8d/700-duplicate-improve-performance)

PR #179, opened 2018-11-09
Fixed flaky resolver tests

SamGodwin2:
Fixes a bug where if the docker image ran slowly a promise in one of the tests would not resolve correctly and would somehow overflow?!? into the next test.

### How to review 
Tests pass even when you only give the docker engine 1 core and 1gb of memory to play with.

### DIFF
--- a/schema/resolvers.test.js
+++ b/schema/resolvers.test.js
@@ -492,7 +492,7 @@ describe("resolvers", () => {
     const parentAnswer = await createNewAnswer(firstPage, "Checkbox");
     const other = await createOther(parentAnswer);
 
-    expect(createOther(parentAnswer)).resolves.toBeNull();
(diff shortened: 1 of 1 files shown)


SamGodwin2:
[Flaky resolvers tests on API](https://trello.com/c/xutHdFfO/714-flaky-resolvers-tests-on-api)

SamGodwin2:
Ah yeah good spot, just found somewhere else we use that `.resolves` without `await`ing the response, I'll update both those tests to bing them in line with what those docs say.

schema/resolvers.test.js
```diff
@@ -588,7 +588,7 @@ describe("resolvers", () => {
       }
     });
 
-    expect(
+    await expect(
```
samiwel: I think you can just return this `expect` since the test function is async.

PR #182, opened 2018-11-14
729 Section intro bug fix

rmccar:
There is a bug with section introductions where you can enter more than 255 characters and wipe the all the text boxes on the page. This is because the columns in the postgress table were of type string and needed to be type text. I have created a new migration to fix this and alter the columns to be of type text

### How to review 
Run tests
Compare with master and make sure bug is now fixed

### DIFF
--- a/migrations/20181114153441_ammended_section_intro_column_types.js
+++ b/migrations/20181114153441_ammended_section_intro_column_types.js
@@ -1,0 +1,20 @@
+const { noop } = require("lodash");
+
+exports.up = async function(knex) {
+  await knex.raw(`DROP VIEW "SectionsView"`);
+  await knex.schema.table("Sections", table => {
+    table.text("introductionTitle").alter();
+    table.text("introductionContent").alter();
+  });
+  return knex.raw(`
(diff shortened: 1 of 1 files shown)


samiwel:
[Section Intro Bug](https://trello.com/c/FlRVCqUZ/729-section-intro-bug)

SamGodwin2 (approved):

migrations/20181114153441_ammended_section_intro_column_types.js
```diff
@@ -0,0 +1,20 @@
+const { noop } = require("lodash");
+
+exports.up = async function(knex) {
+  await knex.raw(`DROP VIEW "SectionsView"`);
+  await knex.schema.table("Sections", table => {
+    table.text("introductionTitle").alter();
+    table.text("introductionContent").alter();
+  });
+  return knex.raw(`
```
tgandrews: Why do we need to recreate the view?

rmccar: We cant amend a table without dropping and recreating the view

tgandrews (approved):
As discussed view block table from being updated.

PR #185, opened 2018-11-16
Tech Debt - Remove legacy watch

tgandrews:
Remove legacy watch. It seems to be resolved by docker-compose build.

This was killing the CPU on my Mac.

### How to review 
1. Checkout this branch
1. Ensure you can change files and the server in docker restarts.

If not.
1. Run `docker-compose build` then `docker-compose up`
1. Then see if it works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "start": "yarn knex -- migrate:latest && nodemon",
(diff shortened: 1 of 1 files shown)


adammahmood:
Amazing. 👍 CPU usage from docker has dropped drastically when API is idle, it only spikes when reloading server.

adammahmood (approved):

